PR #52, opened 2015-01-05
[DISCUSS] Add role and aria-label to notices to improve accessibility

camilleldn:
/cc @tombye @dsingleton

There is currently no way for screen readers to distinguish these notices from the rest of the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ In addition to the [standard Markdown syntax](http://daringfireball.net/projects
 
 creates a callout with an info (i) icon.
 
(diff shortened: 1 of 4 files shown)


bradleywright:
:+1: for intent

camilleldn:
@bradleywright :question: for execution? :-P

bradleywright:
The code looks good too, my only query would be about whether they're all `role=note` or whether there's a better/more specific role to be used for the non-information ones.

Either way it's an improvement.

PR #54, opened 2015-02-18
Fix a small typo in the README

craigw:
There's a small typo in the README that makes it look like Govspeak inserts the word "standard" when talking about VAT in the example `{::highlight-answer}`.


alext:
Thanks for this.

Issue #57, opened 2015-06-26
URI Gotcha with Ruby 2.2.x


jamiecobbett:
Applications using this gem need to take care when upgrading to Ruby 2.2.x as it is less tolerant of non-standard URIs and so new and existing content might no longer be parseable by Govspeak.

More detail:

Ruby 2.2 [upgrades the `URI` module](https://github.com/ruby/ruby/blob/v2_2_0/NEWS#L238-L239) in stdlib to support [RFC3986](https://www.ietf.org/rfc/rfc3986.txt). This is a stricter form of `URI.parse` and means that anywhere we parse URIs that might be invalid, an error will be raised.

Examples found so far include:

* `mailto://address` links (raises `NoMethodError: undefined method 'split' for nil:NilClass`
* blank `mailto:` (raises `NoMethodError: undefined method 'split' for nil:NilClass`)
* URLs with trailing spaces (raises `URI::InvalidURIError: bad URI(is not URI?)`)

This may or may not be something we can/should fix in Govspeak. The alternative is to be aware when upgrading apps using Govspeak to fix existing content.

@boffbowsh did the hard work to figure this out. For more information, see: https://github.com/alphagov/whitehall/pull/2234

tekin:
One option is to use `[Addressable::URI](http://addressable.rubyforge.org/api/Addressable/URI.html)` instead of `URI` - it's less strict when parsing URIs.

h-lame:
Is there any desire to do what @tekin suggests and replace `URI.parse` with `Addressable::URI.parse` to avoid unexpected exceptions?

jamiecobbett:
Fixed in [3.5.1](https://github.com/alphagov/govspeak/pull/63).

PR #59, opened 2015-07-24
Update ruby version for tests to 2.1

alext:
All apps using this are on at least 2.1 now.


jamiecobbett:
Worth broadcasting this gotcha again: https://github.com/alphagov/govspeak/issues/57

PR #60, opened 2015-10-07
Add statistic headline markdown

fofr:
https://trello.com/c/z4YmHWDp/127-html-pubs-add-markdown-to-allow-users-to-add-large-font-and-graphs-medium

> Departments are about to create and publish their Single Departmental Plans. These will be published to GOV.UK as HTML publications. Departments need to be able to pull out important figures periodically throughout the HTML publication in large font.

* Add the `{stat-headline}` markdown pattern
* Add `<aside>` as an acceptable HTML element

See also: https://github.com/alphagov/whitehall/compare/stat-headlines

```
{stat-headline}*13.8bn* years since the big bang{/stat-headline}
```

screen shot 2015-10-08 at 17 01 17

cc @alextea

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,12 +84,12 @@ creates a large highlight box with optional preamble text and giant text denoted
 
 Used in HTML publications.
 
-Statistic headlines highlight important numbers in content. Displays a statistic as a large number with a description. The statistic and description must make sense when read aloud. The giant number is denoted with `**`.
+Statistic headlines highlight important numbers in content. Displays a statistic as a large number with a description. The statistic and description must make sense when read aloud. The important number must be wrapped in `**`.
 
 ```
-{::stat-headline}
+{stat-headline}
 *13.8bn* years since the big bang
-{:/stat-headline}
+{/stat-headline}
 ```
 
 Creates the following:
--- a/README.md
+++ b/README.md
@@ -92,32 +92,12 @@ Statistic headlines highlight important numbers in content. Displays a statistic
 {:/stat-headline}
 ```
 
-You can display many statistic headlines together using a group:
-
-```
-{::stat-headline-group}
-{::stat-headline}*13.8bn* years since the big bang{:/stat-headline}
-{::stat-headline}*100bn* stars in our galaxy{:/stat-headline}
-{::stat-headline}*141m* miles to Mars{:/stat-headline}
-{:/stat-headline-group}
-```
-
 Creates the following:
 
 ```html
-<div class="stat-headline-group">
-  <aside class="stat-headline">
-    <p><em>13.8bn</em> years since the big bang</p>
-  </aside>
-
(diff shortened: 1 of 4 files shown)


jamiecobbett:
cc @dsingleton and @heathd as a person who has worked a lot on Govspeak/Whitehall.

README.md
```diff
@@ -74,12 +74,52 @@ highlights the enclosed text in yellow
...
 
     <div class="highlight-answer">
     <p>The VAT rate is <em>20%</em></p>
     </div>
 
+### Statistic headline
+
+Used in HTML publications.
+
+Statistic headlines highlight important numbers in content. Displays a statistic as a large number with a description. The statistic and description must make sense when read aloud. The giant number is denoted with `**`.
```
jamiecobbett: > The giant number is denoted with `**`.

I can see that is consistent with the `Answer` section, but why is it not something like "wrapped with *'s"?

alextea:
Interested to hear people's thoughts on the govspeak syntax. We have many different styles, and I think it would be good to agree a consensus for any new patterns (and then maybe port existing patterns to new style, and deprecate the old styles).

For the record, I think hyphen delimited text reads better than camel case. Do we need the `::` parts?

heathd:
looks good to me, it's using (one of) the existing patterns.

I agree with @alextea that there's inconsistency, but this isn't making it any worse, cleaning up the inconsistencies will be a bigger piece of work.

dsingleton:
The markup and changes to govspeak itself look good to me :+1: 

Are you planning on adding the [CSS changes in WH](https://github.com/alphagov/whitehall/compare/stat-headlines) to the the [Govspeak component on static](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_govspeak.scss)? It would be good to see a fixture for the Component Guide, to explicitly show  how the `state-headline` will render in that situation?

fofr:
Looking at the `{::stat-headline}` pattern with a few users, it's quickly apparent that it's very easy to make a couple of mistakes:
* `{::` can be confused with `{:`
* `{:/` can be confused with `{/:` or `{::/`

This type of error is then very hard to spot and fix.

fofr:
I've removed `{::stat-headline-group}` in https://github.com/alphagov/govspeak/commit/de962cdac31e66e4256fe0f05b72b16a5c2b816a

fofr:
I've updated the markdown from `{::stat-headline}` to `{stat-headline}` to reduce user error. This adds a new pattern into govspeak but is preferable to the existing one. The previous `{:: }` pattern is not widely known to Whitehall users.

If people are happy with this approach and the new pattern this PR can be merged.

jamiecobbett:
:+1:

dsingleton:
:+1:

PR #63, opened 2015-11-20
Replace URI with Addressable

benlovell:
In order to get past alphagov/govspeak#57 (in summary: ruby 2.2's `URI`
component is strict when parsing particular URIs) we need to switch `URI` for
`Addressable::URI`.

Also clarified copy and intent for `mailto` URI based tests.


jamiecobbett:
Thinking about this a bit more, we should add test cases for the examples described in the [issue](https://github.com/alphagov/govspeak/issues/57).

benlovell:
@jamiecobbett I agree. Addressed in 02b6dd9.

jamiecobbett:
@boffbowsh I've rebased the last commit to add assertions on the output as you suggested.

PR #65, opened 2015-11-23
Test on multiple rubies

jamiecobbett:
We use this gem on multiple Rubies, so we should test on multiple Rubies. Also, the test suite now supports Ruby 2.2.x.


jamiecobbett:
I've just realised that the master build on internal CI won't pay any attention to the travis.yml and so won't test on any other Rubies.

PR #66, opened 2016-01-06
Improve `headers` method

benbarnett:
Fixes two issues to do with nested headers not being picked up, and where explicitly assigned ID's were being ignored in favoured of the auto-generated versions.

**Please be kind with code review – I'm not a ruby developer 😇**

Example:
```md
# Heading one {#explicit-id}

<div markdown="1">
## Heading two
</div>
```

* When calling `headers`, it was only returning the first heading with an incorrect ID (`#heading` as opposed to `#explicit-id`).
* The `<h2>Heading two</h2>` was not being recognised by the govspeak method, even though it was in the rendered HTML.

Follows the kramdown documentation for [Specifying a Header ID](http://kramdown.gettalong.org/syntax.html#specifying-a-header-id) and [HTML Blocks](http://kramdown.gettalong.org/syntax.html#html-blocks).

### DIFF
--- a/lib/govspeak/header_extractor.rb
+++ b/lib/govspeak/header_extractor.rb
@@ -3,11 +3,45 @@ module Govspeak
 
   class HeaderExtractor < Kramdown::Converter::Base
     def convert(doc)
-      doc.root.children.map do |el|
+      headers = []
+
+      doc.root.children.each do |el|
         if el.type == :header
-          Header.new(el.options[:raw_text], el.options[:level], generate_id(el.options[:raw_text]))
+          headers << build_header(el)
+          next
         end
-      end.compact
+
+        headers << find_headers(el) if el.type == :html_element
+      end
+
+      headers.flatten.compact
+    end
+
+  private
+    def id(el)
+      el.attr.fetch('id', generate_id(el.options[:raw_text]))
+    end
+
+    def build_header(el)
+      Header.new(el.options[:raw_text], el.options[:level], id(el))
+    end
+
+    def find_headers(parent)
+      headers = []
+
+      if parent.type == :header
+        headers << build_header(parent)
+      elsif parent.type == :html_element
+        parent.children.each do |child|
+          if child.type == :header
+            headers << build_header(child)
+          elsif child.children.size > 0
+            headers << find_headers(child)
+          end
+        end
+      end
+
+      headers
     end
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ rvm:
   - 1.9.3
   - 2.1
   - 2.2
+before_install:
+  - gem update bundler
 script:
   - bundle exec rake
 branches:
--- a/test/govspeak_test.rb
+++ b/test/govspeak_test.rb
@@ -56,6 +56,43 @@ class GovspeakTest < Minitest::Test
     ], document.headers
   end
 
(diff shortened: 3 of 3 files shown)


lib/govspeak/header_extractor.rb
```diff
@@ -3,11 +3,46 @@ module Govspeak
...
         end
-      end.compact
+
+        headers.push(find_headers(el)) if el.type == :html_element
+      end
+
+      headers.flatten.compact
+    end
+
+    private
```
benlovell: Our style is to indent this at the `class` level and drop the whitespace line beneath. See https://github.com/alphagov/styleguides/blob/master/ruby.md

lib/govspeak/header_extractor.rb
```diff
@@ -3,11 +3,46 @@ module Govspeak
...
+
+    def id(el)
+      el.attr['id'] || generate_id(el.options[:raw_text])
```
benlovell: `el.attr.fetch('id', ...)` would be slightly cleaner.

lib/govspeak/header_extractor.rb
```diff
@@ -3,11 +3,46 @@ module Govspeak
...
+          elsif child.children.size > 0
+            headers.push(find_headers(child))
+          end
+        end
+      end
+
+      headers
     end
   end
-end
```
benlovell: Nice :sparkle:

lib/govspeak/header_extractor.rb
benbarnett: Addressed in https://github.com/guidance-guarantee-programme/govspeak/commit/84c74e6f1d0e05e2ce57fbf12c98f243a7264bc3

lib/govspeak/header_extractor.rb
benbarnett: Addressed in https://github.com/guidance-guarantee-programme/govspeak/commit/84c74e6f1d0e05e2ce57fbf12c98f243a7264bc3

benlovell:
LGTM. Any thoughts @fofr, @dsingleton?

lib/govspeak/header_extractor.rb
```diff
@@ -3,11 +3,45 @@ module Govspeak
 
   class HeaderExtractor < Kramdown::Converter::Base
     def convert(doc)
+      headers = []
+
       doc.root.children.map do |el|
```
boffbowsh: Because you're now explicitly pushing things into the `headers` array, `each` is more appropriate than `map` here. `map` is used to create an array of the items that the iterator block yields, whereas `each` just executes the block once per item.

lib/govspeak/header_extractor.rb
```diff
@@ -3,11 +3,45 @@ module Govspeak
...
         if el.type == :header
-          Header.new(el.options[:raw_text], el.options[:level], generate_id(el.options[:raw_text]))
+          headers.push(build_header(el))
```
boffbowsh: Personal preference maybe, but I'd prefer `headers << build_header(el)`

boffbowsh:
Couple of tiny comments, otherwise :+1: 

> Please be kind with code review – I'm not a ruby developer 😇
You are now :smile:

benbarnett:
Thanks both, comments should all be addressed now!

benbarnett:
@boffbowsh @benlovell either of you free to push the button?

PR #69, opened 2016-02-23
[DON'T MERGE BUT HELP!] Add failing test for legislative list bug [DON'T SEAL]

binaryberry:
When a footnote is inserted in a legislative list, it does not
get processed by govspeak. Other govspeak does get processed, but
footnotes are not standard markdown http://kramdown.gettalong.org/syntax.html#footnotes.

We tried upgrading the kramdown gem to 1.9.0 but that didn't make the
test pass.

### DIFF
--- a/test/govspeak_test.rb
+++ b/test/govspeak_test.rb
@@ -569,6 +569,24 @@ $CTA
     }
   end
 
+  test_given_govspeak "This bit of text\r\n\r\n$LegislativeList\r\n* 1. has a footnote[^1]\r\n$EndLegislativeList\r\n[^1]: This is a footnote." do
+    assert_html_output %{
+      <p>This bit of text</p>
+
+      <ol class="legislative-list">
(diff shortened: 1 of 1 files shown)


jamiecobbett:
I have simplified the steps to reproduce down to:
```ruby
text = %{
<p>Something about this form download[^1]</p>


[^1]: And here is the definition.
}
Kramdown::Document.new(text).to_html
```

It's something to do with a `<` following the footnote marker. It seems to process successfully if it is followed by a full stop, whitespace or an HTML entity (`&nbsp;`).

elliotcm:
@binaryberry Do we still need this PR?

binaryberry:
Hi @elliotcm, unfortunately I'm unable to find the related zendesk ticket to check its current status (I'm kicking myself for not linking to it in the PR), but I believe I had helped the user with a workaround and passed on the proper bugfix to product triage with a mention of this PR. We can probably close for now.

elliotcm:
OK

PR #71, opened 2016-04-29
Allow table cells to have a text alignment

fofr:
Kramdown has a markdown pattern for allowing left, centred and right aligned cell contents:
http://kramdown.gettalong.org/quickref.html#tables

To achieve this it uses a style property, eg `style=‘text-align: right’`
https://github.com/gettalong/kramdown/blob/60795eb95f1588d41e105407add2d3ba6a7fc101/lib/kramdown/parser/html.rb#L446-L449

We don’t want to allow any styles to be applied, just the subset that Kramdown uses. Instead, whitelist the attribute for table cells, but use a Sanitise transformer to remove the style unless it matches the 3 possible values it could have.

Part of:
https://trello.com/c/6ld11Eck/346-right-alignment-in-tables-in-govspeak-medium

Markdown example (left aligned, centre aligned, right aligned):
```
| Header1 | Header2 | Header3 |
|:--------|:-------:|--------:|
| cell1   | cell2   | cell3   |
| cell4   | cell5   | cell6   |
```

| Header1 | Header2 | Header3 |
|:--------|:-------:|--------:|
| cell1   | cell2   | cell3   |
| cell4   | cell5   | cell6   |

### DIFF
--- a/lib/govspeak/html_sanitizer.rb
+++ b/lib/govspeak/html_sanitizer.rb
@@ -21,13 +21,26 @@ class Govspeak::HtmlSanitizer
     end
   end
 
+  class TableCellTextAlignWhitelister
+    def call(sanitize_context)
+      return unless ["td", "th"].include?(sanitize_context[:node_name])
+      node = sanitize_context[:node]
+
+      # Kramdown uses text-align to allow table cells to be aligned
+      # http://kramdown.gettalong.org/quickref.html#tables
+      unless node['style'].match(/^text-align:\s*(center|left|right)$/)
+        node.remove_attribute('style')
+      end
+    end
+  end
+
   def initialize(dirty_html, options = {})
     @dirty_html = dirty_html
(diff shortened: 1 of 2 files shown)


fofr:
An alternative to letting these styles through might be to transform Kramdown's output to use an `align` attribute. Github's rendering uses align. We'd also need to add the styles to correctly align the content.

dsingleton:
> An alternative to letting these styles through might be to transform Kramdown's output to use an align attribute. Github's rendering uses align. We'd also need to add the styles to correctly align the content.

I don't think the `align` attribute would need styling ([demo](http://jsbin.com/tilibazuxi/edit?html,output)), but it's been dropped in HTML 5.

I'm not a big fan of the inline styles, we could use `align` regardless, or alignment classes?

fofr:
We should raise an issue upstream on Kramdown.

PR #73, opened 2016-05-03
Allow td and th elements without a style attribute

fofr:
If no `style` was included `node['style'].match` returned a no method error, so normal tables wouldn’t work correctly.


boffbowsh:
Merge when 💚

PR #74, opened 2016-05-03
Version 3.6.1

fofr:
Yank version 3.6.0 because of #73 (td and th elements without a style attribute would error)


boffbowsh:
Merge when 💚

PR #78, opened 2016-07-07
[Discuss] Whitehall fractions spike

elliotcm:
We'd like to move whitehall's custom govspeak extensions into the gem, and two questions were raised:
1. How do we handle assets (images for fractions, JS for sortable tables, styles for a number of extensions)
2. How do we check to confirm that the output of the gem matches the original output from whitehall

Embedding JS and CSS into HTML is already easy using `script` and `style` tags, so I looked into embedding images via data URIs.  The images for fractions are quite small, so they're a good candidate for this kind of embedding.

The attached code implements one of the fractions.  Here's a comparison:

Before (loaded from our assets host):
screenshot 2016-07-07 11 27 50

After (embedded in the HTML):
screenshot 2016-07-07 11 28 00

The small images come in way under 32kB so this approach is compatible as far back as IE8 according to http://caniuse.com/#feat=datauri

As for checking the output, I'm not sure how much of an issue that is.  It would be possible with some effort to have our async experiments or github scientist check the output of app helpers and govspeak gem to make sure they match, but I think we can do the checks manually.

The styles and scripts for the extensions are fixed, so it seems like normal sanity checks would be enough to confirm it's working correctly.

The images for the fractions change each time but there's a fixed number of them (32) so it would be easy enough to create a page with all of them on and confirm they look right visually.

cc @danielroseman @dsingleton 

https://trello.com/c/W7m0nc3P/747-prepare-to-move-whitehall-govspeak-functionality-into-gem


elliotcm:
Closing this as it's a discussion thread not intended to be merged.

dsingleton:
Initial thoughts:
- Using `datauri` feels like the neatest option, it's pretty self contained, and understandable
- But support for older versions of IE may be a blocker. The images are content, not enhancement, the visual style of the fractions is important to the content, and not displaying may mislead readers. This needs some thought, and probably looking at older-IE usage on this document.
- In this approach should the source images live _in_ the gem? Or should they be passed in as a list of available fractions for that invocation? Currently fractions are a pretty niche feature for a single

lib/govspeak/fractions.rb
```diff
@@ -0,0 +1,11 @@
...
...y5hZG9iZS5jb20veGFwLzEuMC8iIHhtbG5zOmRjPSJodHRwOi8vcHVybC5vcmcvZGMvZWxlbWVudHMvMS4xLyIgeG1wTU06T3JpZ2luYWxEb2N1bWVudElEPSJ1dWlkOjY1RTYzOTA2ODZDRjExREJBNkUyRDg4N0NFQUNCNDA3IiB4bXBNTTpEb2N1bWVudElEPSJ4bXAuZGlkOjkxNTc3NjYxMEU4QTExRTNCRUE0QjBBQUNBQUYwNDI4IiB4bXBNTTpJbnN0YW5jZUlEPSJ4bXAuaWlkOjkxNTc3NjYwMEU4QTExRTNCRUE0QjBBQUNBQUYwNDI4IiB4bXA6Q3JlYXRvclRvb2w9IkFkb2JlIElsbHVzdHJhdG9yIENTNiAoTWFjaW50b3NoKSI+IDx4bXBNTTpEZXJpdmVkRnJvbSBzdFJlZjppbnN0YW5jZUlEPSJ4bXAuaWlkOjRBNjMzQzkzODQyMTY4MTE4MDgzQjk5QzFCRkRFNEI5IiBzdFJlZjpkb2N1bWVudElEPSJ4bXAuZGlkOjA1ODAxMTc0MDcyMDY4MTE4M0QxRUZCQkU2ODE5RTYxIi8+IDxkYzp0aXRsZT4gPHJkZjpBbHQ+IDxyZGY6bGkgeG1sOmxhbmc9IngtZGVmYXVsdCI+RnJhY3Rpb25zPC9yZGY6bGk+IDwvcmRmOkFsdD4gPC9kYzp0aXRsZT4gPC9yZGY6RGVzY3JpcHRpb24+IDwvcmRmOlJERj4gPC94OnhtcG1ldGE+IDw/eHBhY2tldCBlbmQ9InIiPz4qxAJzAAABEUlEQVR42mJmQAVsQCwHxD+B+A8DFpALxO+A+BkQfwfiqUDMjKxAFyrhCOWDTHsNxHEgDgtU8DIQywLxGyj/ERCfA2IZBhxAE4izgPgVEKvjUgRy018gngb1CE4gBcRXgbgCxGFCEnRGM20fEBsi6/QD4h9AHAjE7EBsBMRPgbgI3YoUIL4DxP+B+AUQd6CHE50AI9QNgxXsB+IHaNgN5nAYAKUnGyB+iyQGSgnfYBwhIP4FZXMipTMG5AiWhsbdWSB+D8XV6O4BJa4VQGwOdYI1EH8G4hBCHpkPxIuQrQPp5kdTBEqd/5AFMqDJQx7KVwHiD0AciqwI5Jt5QPwViG9Ag6MLlzuEgdgASsMBQIABACcRNstEO18eAAAAAElFTkSuQmCC"
```
cbaines: Where did this come from, and how was it generated?

PR #80, opened 2016-08-03
Govspeak support for attachment

dougdroper:
This moves the logic from whitehall into govspeak.
see:
https://github.com/alphagov/whitehall/blob/master/app/views/documents/_attachment.html.erb

https://trello.com/c/K4Rq6Sag/767-move-attachments-support-into-govspeak-gem-medium


test/govspeak_test.rb
```diff
@@ -827,4 +827,39 @@ class GovspeakTest < Minitest::Test
...
+      opendocument?: true,
+      order_url: "www.test.com",
+      price: 12.3,
+      csv?: true,
+      unnumbered_command_paper?: true,
+      isbn: 'isbn-123',
+    )
+    govspeak = "[embed:attachments:2b4d92f3-f8cd-4284-aaaa-25b3a640d26c]"
+    rendered = Govspeak::Document.new(govspeak, {attachments: attachment}).to_html
+    assert_match /<section class=\"attachment embedded\">/, rendered
```
kevindew: This assert and the ones below all trigger a ruby warning when you run the tests: http://stackoverflow.com/questions/5239805/what-triggers-the-ruby-warning-about-ambiguous-first-argument

can silence it by either using parenthesis on that method call or `%r{}` syntax

lib/presenters/attachment_presenter.rb
```diff
@@ -0,0 +1,233 @@
...
+    attachment_info << "  Original format: #{attachment.file_extension}"
+    attachment_info << "  ISBN: #{attachment.isbn}" if attachment.isbn.present?
+    attachment_info << "  Unique reference: #{attachment.unique_reference}" if attachment.unique_reference.present?
+    attachment_info << "  Command paper number: #{attachment.command_paper_number}" if attachment.command_paper_number.present?
+    if attachment.hoc_paper_number.present?
+      attachment_info << "  House of Commons paper number: #{attachment.hoc_paper_number}"
+      attachment_info << "  Parliamentary session: #{attachment.parliamentary_session}"
+    end
+
+    mail_to alternative_format_contact_email, alternative_format_contact_email,
```
kevindew: Parenthesis on this mail_to might make this method call easier to understand

dougdroper: yeah, parenthesis is the least of my issues with this method. 

Did a brief clean up though to *try* reduce the noise. Still this whole thing is pretty messy. We will almost certainly need to clean it up when we use it in real life. This was mostly a blind copy from whitehall job.

PR #81, opened 2016-08-05
Support Whitehall blockquote changes

kevindew:
This PR introduces two features of whitehall related to blockquotes, both are somewhat interesting and potentially contentious.

### 1. Removing quotes from blockquotes
These are removed as quotes are added automatically as part of styling on the frontend. It will change markdown of `> "test"` to `> test`. 

As this was a global applying of a function I placed it outside the extension system as it didn't have a regexp to match - this doesn't feel a very nice way to handle it though, so I welcome suggestions on how to improve this.

This also may not work particularly well. I've used the same code as Whitehall so it has full compatibility, but it does look like there are some dodgy ways it can run.

### 2. adding a class to the last `<p>` element within a blockquote. 

For example:
```
<blockquote>
  <p>First Paragraph</p>
  <p>Second Paragraph</p>
</blockquote>
```
becomes:
```
<blockquote>
  <p>First Paragraph</p>
  <p class="last-child">Second Paragraph</p>
</blockquote>
```
To achieve this we've used nokogiri as per the implementation on Whitehall. This has had a number of side effects:

1. Formatting of the HTML returned is different (notably block level elements are on new lines)
2. HTML encoded UTF-8 characters that nokogiri considers safe are output as UTF-8 rather than HTML encoded. For example: `&yen;` is returned as `¥`
3. HTML encoding done numerically is converted to symbolically. For example `&#62;` is returned as `&gt;`

For this feature I've added a PostProcessor class which has a similar extension system to that of the extensions within the main govspeak class, a key difference however being that these are performed on the entire body rather than a matched section. It has the potential though that this could be changed to match a css selector to scope the nokogiri elements coming in.


lib/govspeak/blockquote_extra_quote_remover.rb
```diff
@@ -0,0 +1,19 @@
...
+    # used to remove quotes from a markdown blockquote, as these will be inserted
+    # as part of the rendering
+    #
+    # for example:
+    # > "test"
+    #
+    # will be formatted to:
+    # > test
+    def self.remove(source)
+      return nil if source.nil?
```
dougdroper: nil is redundant here

test/blockquote_extra_quote_remover_test.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  test "ignores text without double quotes" do
+    assert_leaves_untouched %{no quotes\na few lines,\n\n\n\nbut no quotes\n\n}
+  end
+
+  test "ignores text without a quote symbol" do
+    assert_leaves_untouched %{quotes\n"but not with a > symbol"}
+  end
+
+  # TODO check this is actually behaviour we want
```
dougdroper: who do we need to check with?

lib/govspeak/post_processor.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def output
+      document = nokogiri_document
+      @@extensions.each do |_, block|
+        instance_exec(document, &block)
+      end
+      document.to_html
+    end
+
+    extension("add class to last p of blockquote") { |document|
```
dougdroper: can we use `do`..`end` rather then `{}`

test/blockquote_extra_quote_remover_test.rb
kevindew: I'm not sure really - I could probably phrase this note better to indicate that it's kept for consistency but this may not be desired behaviour into the future

dougdroper: I guess it might not be the behaviour that they want, but it is the behaviour that they have been getting :) 

I would remove the comments.

PR #85, opened 2016-08-11
Update addressable version to 2.4.0

mgrassotti:
Trello https://trello.com/c/WLaSMFp0/450-upgrade-govuk-content-api-ruby-dependency-small

It is needed to update ruby version to 2.3.0 on govuk_content_api.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.7.0
+
+* Update Addressable version from 2.3.8 to 2.4.0
+
 ## 3.6.2
 
 * Fix bug with link parsing introduced in Kramdown 1.6.0 with the "inline attribute lists" feature which clashed with our monkey patch [#75](https://github.com/alphagov/govspeak/pull/75)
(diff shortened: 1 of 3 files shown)


boffbowsh:
Looking at the [CHANGELOG](https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md) for addressable, it looks like 2.4.0 introduces some additional strictness which might break things. Perhaps rather than forcing an upgrade to 2.4.0 in here, the requirement could be relaxed? I think we do require at least 2.3.8, but maybe we do something like `[">= 2.3.8", "< 3"]`:

```ruby
irb(main):013:0> r = Gem::Requirement.new([">= 2.3.8", "< 3"])
irb(main):014:0> v238 = Gem::Version.new("2.3.8")
irb(main):015:0> v240 = Gem::Version.new("2.4.0")
irb(main):016:0> v3 = Gem::Version.new("3.0.0")
irb(main):017:0> r =~ v238
=> true
irb(main):018:0> r =~ v240
=> true
irb(main):019:0> r =~ v3
=> false
```

andrewgarner:
@boffbowsh are you happy with this now?

dougdroper:
Just so that you are aware.. there are a few other changes in master: https://github.com/alphagov/govspeak/compare/v3.6.2...master
And we may need to change the ruby-version to 2.3.0.

PR #87, opened 2016-09-21
CLI rendering via govspeak binary

kevindew:
This adds in a binary so that you can render govspeak from the command line, hardly an essential feature but it's been very useful in debugging how govspeak interacts. 

I figure it may also be useful if there are any non ruby applications that need to make use of govspeak.

## Usage

- As an argument: `govspeak "render me"`
- From a file: `govspeak --file render-me.md`
- From stdin: `echo "render-me" | govspeak`
- Via bundler: `bundle exec govspeak "render-me"`

Outputs to stdout

Govspeak options can be passed in as JSON as a string through`--options` or as a file through `--options-file options.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,18 @@ then create a new document
     doc = Govspeak::Document.new "^Test^"
     puts doc.to_html
 
+or alternatively, run it from the command line
+
+    $ govspeak "render-me"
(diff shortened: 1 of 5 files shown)


kevindew:
just spotted an error that I'll sort

PR #89, opened 2016-09-27
Use hashes as options for govspeak extensions

kevindew:
This changes the expected input for our govspeak extensions from objects to hashes. 

This makes sense as it reduces the need for applications that use govspeak to create classes that conform to interfaces defined here and since hash is the format we receive options in the publishing API we don't need to convert that to a class before passing to govspeak. It also maintains our ability to render govspeak with JSON.

I was able to change the options for links, attachments, inline attachments and contacts as we know these aren't used. However we still have [inline attachment images which require a class](https://github.com/alphagov/govspeak/blob/91898938ea9a4edb4a0227aa90464f5b28ae749a/lib/govspeak.rb#L180-L188) I'm not sure if anyone uses that and if we can remove or deprecate it?

Along the way through this I found there were a number of issues with our rendering of attachments where they were getting malformed so I've put together more thorough tests for them and fixed the issues I could fix.

There are still a few issues that should be looked into when the attachments code is started to be rolled out:

- The images referenced in an attachment are unlikely to resolve as they expect to be at `/images/*.png` and I assume they need to be pointed at some host
- We have English in the presenter which would be echo’d in different locales, this appears to be a problem in Whitehall currently

I’ve left FIXME notes in the code for them, I don't see them as high priority as we'll probably have to resolve them as well as others once apps that use attachments start being migrated.


lib/presenters/contact_presenter.rb
```diff
@@ -0,0 +1,74 @@
+require 'active_support/core_ext/array'
+
+class ContactPresenter
+  attr_reader :contact
+
+  def initialize(contact)
+    @contact = contact
```
carlosmartinez: to avoid all these methods wrapping around the `contact` hash, maybe store `contact` as an openstruct (forgive me, mama) and use `delegate` to access them?

carlosmartinez:
approve12

PR #95, opened 2016-10-03
Duplicate options passed into Govspeak

kevindew:
Rather than mutating as we alter it in the initialize method

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -32,7 +32,7 @@ module Govspeak
     end
 
     def initialize(source, options = {})
-      options.deep_symbolize_keys!
+      options = options.dup.deep_symbolize_keys
       @source = source ? source.dup : ""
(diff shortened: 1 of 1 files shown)


lib/govspeak.rb
```diff
@@ -32,7 +32,7 @@ def self.to_html(source, options = {})
     end
 
     def initialize(source, options = {})
-      options.deep_symbolize_keys!
+      options = options.dup.deep_symbolize_keys
```
carlosmartinez: doesn't this also mutate `options`, since you're assigning to it?

kevindew: Nah, the assignment doesn't mutate the original object. As that assignment has no effect outside of this lexical scope. Not sure I explained this well though.

carlosmartinez: cool, thanks

PR #99, opened 2016-10-09
Loosen ActionView dep for Rails 5 compatibility.

danielroseman:
Since Travis builds this gem on different Ruby versions, and ActionView 5 won't install on versions before 2.3, we use a conditional expression in the gemspec to ensure it installs the right one.

### DIFF
--- a/govspeak.gemspec
+++ b/govspeak.gemspec
@@ -32,7 +32,7 @@ library for use in the UK Government Single Domain project}
   s.add_dependency "sanitize", "~> 2.1.0"
   s.add_dependency 'nokogiri', '~> 1.5'
   s.add_dependency 'addressable', '>= 2.3.8', '< 3'
-  s.add_dependency 'actionview', '~> 4.1'
+  s.add_dependency 'actionview', '>= 4.1', '< 6'
   s.add_dependency 'i18n', '~> 0.7'
   s.add_dependency 'money', '~> 6.7'
   s.add_dependency 'commander', '~> 4.4'
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.gem
 .bundle
 Gemfile.lock
+gemfiles/*.lock
 pkg/*
 .DS_Store
 *.pyc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 language: ruby
 sudo: false
-rvm:
-  - 2.1
-  - 2.2
-  - 2.3.1
+matrix:
+  include:
+  - rvm: 2.1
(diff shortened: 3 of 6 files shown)


govspeak.gemspec
```diff
@@ -32,7 +32,7 @@ library for use in the UK Government Single Domain project}
   s.add_dependency "sanitize", "~> 2.1.0"
   s.add_dependency 'nokogiri', '~> 1.5'
   s.add_dependency 'addressable', '>= 2.3.8', '< 3'
-  s.add_dependency 'actionview', '~> 4.1'
+  s.add_dependency 'actionview', RUBY_VERSION >= '2.3' ? '>= 4.1' : '~> 4.1'
```
kevindew: Nice workaround 👍 . It might be prudent to add an upper bound of `'< 6'`

danielroseman: Good point, done.

boffbowsh:
This will set a dependency on the version of Ruby that is used on rubygems.org to generate the dependencies for Bundler. Also, if the only dependency is `deep_symbolize_keys`, we should be depending on `activesupport` not `actionview`.

Is this change for build and test time, or for production?

danielroseman:
@boffbowsh I'm not sure I understand the distinction - that may be because I don't know what rubygems.org is doing. The point is that I want to be able to use this gem in apps that use Rails 4 or 5. Rails 5 won't even install on earlier Ruby versions, but that shouldn't matter; as I understood it, the dependency graph is calculated when you do `bundle install` in an app that uses the gem.

(And it turns out that we do use more than `deep_symbolize_keys`; we include some of the ActionView helpers in the AttachmentPresenter. I'll update the comment.)

boffbowsh:
In theory, yes, but Bundler precalcs the dependency tree and generates an endpoint like this: https://index.rubygems.org/info/govspeak. Using Ruby to [non-]determine which versions are available means that endpoint will be unreliable.

Could we make this something like `">= 4.1", "< 6"` instead?

danielroseman:
That's basically what I had originally, but then Travis failed to build the gem for Ruby 2.1 and 2.2, see https://travis-ci.org/alphagov/govspeak/builds/164866565. I think there are still apps on those versions that use this, so we probably do want to test against them.

boffbowsh:
I think we get around this in other gems by using different Gemfiles for testing: https://github.com/alphagov/slimmer/tree/master/gemfiles

danielroseman:
@boffbowsh does this look right now?

boffbowsh:
You'll still need a dependency in the gemspec so that apps know to install actionview. `">= 4.1", "< 6"` should be sufficient there

danielroseman:
Done.

PR #105, opened 2017-05-03
Convert regex matches to equals

kevindew:
This changes the usage of match method on string to be use of the
equality operator. As the contents is user input it should not be
treated as a regex and there shouldn't be scenarios where a previous
match is no longer a match.

This is to fix an error that was occurring of: "premature end of
char-class: /1-1992 G-BJRT[2/"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fix matching links/attachments/contacts by regex to use equality
+
 ## 5.0.2
 * Loosen ActionView dependency to allow use with Rails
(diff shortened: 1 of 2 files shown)


bevanloon:
just a quick one on the description for the PR....

"match method on string **to be use of the** equality operator."
to be or to use....I'm sure there's an existential joke in there somewhere :)

kevindew:
ha whoops I'll sort that

fofr:
Any chance a failing test/test could be added to demonstrate the issue with `"premature end of
char-class: /1-1992 G-BJRT[2/"`?

PR #108, opened 2017-05-04
Build on Jenkins 2

kevindew:
We were no longer having builds run by our CI which meant that releases were no longer automatic.

Since our CI will run on every commit I've moved the tests that travis was doing to be within our build pipeline, so that we can't accidentally ship a gem where the build fails.


Jenkinsfile
```diff
@@ -0,0 +1,64 @@
...
+
+        govuk.rubyLinter("bin lib test")
+
+        govuk.runTests()
+
+        publishHTML(target: [
+          allowMissing: false,
+          alwaysLinkToLastBuild: false,
+          keepAll: true,
+          reportDir: 'coverage/rcov',
```
thomasleese: Mixed single and double quotes.

kevindew: Thanks! it was a total mess on that front (insert some sort of shame emoji)

PR #118, opened 2017-11-21
Fix buttons rendering incorrectly in linebreaks

nickcolley:
When testing with govspeak-preview I found buttons did not render as expected when using the preview mode,
even though they render fine when pulled in as a markdown file.

This ensures that these edge cases are considered and includes a test.

http://rubular.com/r/t7k2i3iwRd

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -136,7 +136,7 @@ module Govspeak
     end
 
     extension('button', %r{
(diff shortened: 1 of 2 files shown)


nickcolley:
I'm going to try and test this on publisher integration...

nickcolley:
screen shot 2017-11-21 at 16 45 45

PR #120, opened 2018-01-04
Add LinkExtractor class

camillavk:
This LinkExtractor class is used in several different publishing applications when checking links on documents. It extracts any links on the document, excluding ones starting with 'mailto:' or '#', and returns an array of the links present.

### DIFF
--- a/lib/govspeak/link_extractor.rb
+++ b/lib/govspeak/link_extractor.rb
@@ -1,0 +1,28 @@
+module Govspeak
+  class LinkExtractor
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      @links ||= extract_links
+    end
+
+  private
+
+    def extract_links
+      document_anchors.map { |link| link['href'] }
+    end
+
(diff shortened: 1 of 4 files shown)


lib/govspeak/link_extractor.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    def call
+      @links ||= extract_links
+    end
+
+  private
+
+    def extract_links
+      processed_govspeak.css('a:not([href^="mailto"])').css('a:not([href^="#"])').map { |link| link['href'] }
```
thomasleese: I would maybe extract this line out to another method so it's something like:

```ruby
def extract_links
  document_anchors.map { |link| link['href'] }
end

def document_anchors
  processed_govspeak.css('a:not([href^="mailto"])').css('a:not([href^="#"])')
end
```

camillavk:
@thomasleese I've updated the CHANGELOG :)

PR #122, opened 2018-01-11
Add website_url arguement to Govspeak::Document#extracted_links

SebAshton:
This PR adds the ability to set an optional argument on `Govspeak::Document#extracted_links` which will mean all returned URLs are fully qualified.


test/govspeak_link_extractor_with_website_root_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+
+class GovspeakLinkExtractorWithWebsiteRootTest < Minitest::Test
```
thomasleese: I would be tempted to just merge this in with the other tests for the link extractor.

PR #131, opened 2018-11-05
Bring gem up to date with other GOV.UK gems

kevindew:
Loosely related trello: https://trello.com/c/gEAMmD0g/386-ability-to-embed-a-contact-via-markdown

As there hasn't been active development on this gem for a while it's pretty out of date with other GOV.UK gems.

I've updated this so:

- There aren't any linting issues
- Use `govuk.buildProject` to build it (which now tests it against rubies 2.3, 2.4 and 2.5 rather than 2.1, 2.2 and 2.3)
- Drop support for Rails 4
- Remove an unreliable test that produces different results based on ActiveSupport version
- Stop getting deprecation warnings when using Money gem

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,64 +2,6 @@
 
 library("govuk")
 
-REPOSITORY = "govspeak"
-
-def rubyVersions = [
-  "2.1",
-  "2.2",
-  "2.3.1",
-]
-
 node {
-
-  try {
-    stage("Checkout") {
-      checkout(scm)
-      govuk.mergeMasterBranch()
-    }
-
-    for (rubyVersion in rubyVersions) {
-      stage("Test with ruby $rubyVersion") {
-        govuk.cleanupGit()
-        govuk.setEnvar("RBENV_VERSION", rubyVersion)
-        govuk.setEnvar("BUNDLE_GEMFILE", "gemfiles/Gemfile.ruby-${rubyVersion}")
-        govuk.bundleGem()
-
-        govuk.rubyLinter("bin lib test")
-
-        govuk.runTests()
-
-        publishHTML(target: [
-          allowMissing: false,
-          alwaysLinkToLastBuild: false,
-          keepAll: true,
-          reportDir: "coverage/rcov",
-          reportFiles: "index.html",
-          reportName: "RCov Report ${rubyVersion}"
-        ])
-      }
-    }
-    sh("unset RBENV_VERSION")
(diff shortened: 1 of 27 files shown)


govspeak.gemspec
```diff
@@ -28,7 +28,7 @@ library for use in the UK Government Single Domain project'
   s.executables   = s.files.grep(%r{^bin/}) { |f| File.basename(f) }
   s.require_paths = %w[lib]
 
-  s.add_dependency 'actionview', '>= 4.1', '< 6'
+  s.add_dependency 'actionview', '~> 5.0'
```
benthorner: I'm assuming you know we're at 5 everywhere we need to be ;-). (Or this forces an upgrade.)

test/govspeak_attachments_test.rb
```diff
@@ -344,59 +344,6 @@ def render_govspeak(govspeak, attachments = [], options = {})
     end
   end
 
-  test "a full attachment rendering looks correct" do
```
benthorner: I see what you mean - there are lots of tests in this file already that seem to be checking the same things :-).

Jenkinsfile
```diff
@@ -2,64 +2,6 @@
...
-
-  } catch (e) {
-    currentBuild.result = "FAILED"
-    step([$class: "Mailer",
-          notifyEveryUnstableBuild: true,
-          recipients: "govuk-ci-notifications@digital.cabinet-office.gov.uk",
-          sendToIndividuals: true])
-    throw e
-  }
+  govuk.buildProject(rubyLintDiff: false)
```
benthorner: Didn't realise govuk.groovy was doing so many clever little things! This looks similar from a cursory glance :-).

Rakefile
```diff
@@ -12,10 +12,4 @@ Rake::TestTask.new("test") { |t|
...
 }
 
-require "gem_publisher"
-task :publish_gem do |t|
-  gem = GemPublisher.publish_if_updated("govspeak.gemspec", :rubygems)
-  puts "Published #{gem}" if gem
-end
-
-task :default => [:test]
+task default: [:test]
```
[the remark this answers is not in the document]
kevindew: It needs some more configuration for lint to be in there as well to load the task in.

I also don't think we have a defined pattern for rake default including linting although I do agree it's a good thing as we then have to override the test task which is an exceptional thing. Since this repo doesn't get much love I'm not too keen to take it down an exceptional route.

PR #136, opened 2018-12-07
New image extension

benthorner:
https://trello.com/c/9MSu3fpP/532-add-support-for-imagefilenamepng-to-govspeak

This adds support for a [Image:image-id] syntax to deprecate the old !!n syntax, both of which make use of an 'images' array but expect different data structures:

   * The old data structure is a method with :url, :alt_text and :caption methods
   * The new data structure is a hash with :url, :alt_text, :caption and :id keys

Using the old syntax with the new data structure will work, while using the new syntax with the old data structure will not. Any form of ID can be used for the new data structure, although a filename is recommended for readability.

### DIFF
--- a/lib/govspeak/presenters/image_presenter.rb
+++ b/lib/govspeak/presenters/image_presenter.rb
@@ -7,15 +7,15 @@ module Govspeak
     end
 
     def url
-      image.url
+      image[:url]
     end
 
     def alt_text
-      image.alt_text
+      image[:alt_text]
     end
 
     def caption
-      image.caption.strip.presence rescue nil
+      image[:caption].to_s.strip.presence
     end
 
     def id
--- a/lib/govspeak/presenters/image_presenter.rb
+++ b/lib/govspeak/presenters/image_presenter.rb
@@ -1,0 +1,25 @@
+module Govspeak
+  class ImagePresenter
+    attr_reader :image
+
+    def initialize(image)
+      @image = image
+    end
+
+    def url
+      image.url
+    end
+
+    def alt_text
+      image.alt_text
+    end
+
+    def caption
+      image.caption.strip.presence rescue nil
+    end
+
+    def id
+      nil
+    end
+  end
+end
--- a/lib/govspeak/presenters/image_presenter.rb
+++ b/lib/govspeak/presenters/image_presenter.rb
@@ -7,15 +7,15 @@ module Govspeak
     end
 
     def url
-      image[:url]
+      image.respond_to?(:url) ? image.url : image[:url]
     end
 
     def alt_text
-      image[:alt_text]
+      image.respond_to?(:alt_text) ? image.alt_text : image[:alt_text]
     end
 
     def caption
-      image[:caption].to_s.strip.presence
+      (image.respond_to?(:caption) ? image.caption : image[:caption]).to_s.strip.presence
     end
 
     def id
--- a/test/govspeak_images_test.rb
+++ b/test/govspeak_images_test.rb
@@ -1,0 +1,60 @@
+# encoding: UTF-8
+
+require 'test_helper'
+require 'govspeak_test_helper'
+
+class GovspeakImagesTest < Minitest::Test
+  include GovspeakTestHelper
+
+  test "can reference attached images using !!n" do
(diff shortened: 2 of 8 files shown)


lib/govspeak/presenters/image_presenter.rb
```diff
@@ -0,0 +1,25 @@
...
+    def initialize(image)
+      @image = image
+    end
+
+    def url
+      image.respond_to?(:url) ? image.url : image[:url]
+    end
+
+    def alt_text
+      image.respond_to?(:alt_text) ? image.alt_text : image[:alt_text]
```
kevindew: Do we need to do any of the new line trimming that's done in AttachmentImagePresenter? I don't know off the top of my head what issue they were resolving there.

lib/govspeak/presenters/image_presenter.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def caption
+      (image.respond_to?(:caption) ? image.caption : image[:caption]).to_s.strip.presence
+    end
+
+    def id
+      nil
```
kevindew: Does this method need to exist?

test/govspeak_images_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  test "Image:image-id syntax renders nothing if not found" do
+    doc = Govspeak::Document.new("[Image:another-id]")
+    assert_equal %{\n}, doc.to_html
+  end
+
+  test "Image:image-id syntax adds image caption if given" do
+    given_govspeak "[Image:image-id]", images: [build_image(caption: 'My Caption & so on')] do
```
kevindew: Do we know whether the whitehall caption equivalent supports or escapes HTML and that this is equivalent (hopefully it escapes)

lib/govspeak/presenters/image_presenter.rb
benthorner: I don't think so: partly as we're instead copying Whitehall which seems not to bother (although it might do it in the app); and partly as we know we've got a plain input.

lib/govspeak/presenters/image_presenter.rb
benthorner: This is called by the `render_image` method, so the presenter has to have the same interface. It is defined in AttachmentImagePresenter, and it seems reasonable that we might want to define it properly for images, someday i.e. the common interface seems reasonable.

test/govspeak_images_test.rb
benthorner: I don't think it does - it just calls directly with edition.images, and there doesn't seem to be any escaping on input. But surely ERB will do that automatically?

lib/govspeak/presenters/image_presenter.rb
kevindew: Ah gotcha 👍

lib/govspeak/presenters/image_presenter.rb
kevindew: ok cool

test/govspeak_images_test.rb
kevindew: Rails does it rather than ERB if I remember correctly. as this uses active support I'm not sure if ERB would be patched or not but I expect this would be vulnerable to outputting HTML (you can see the encode method on alt text and filename)

benthorner: Oops, forgot about the fun-filled textual render method ;-). No ERB to help there anyway - I'll add an encode to that as part of this.

kevindew: cool probably worth doing a cursory check that Whitehall doesn't pass govspeak through or anything - certainly hope not.

benthorner: Oh, that's an idea - but I tried creating a document with an image like that and the govspeak caption isn't processed. Makes some sense, actually, as otherwise you could have recursive govspeak extensions generating their own govspeak.

benthorner:
I forgot to mention that in the description - I don't think anyone's going to look at the code to see it's deprecated, but it turns out no one ever wrote about the !!n approach in the README. We could add a section to specifically describe it as deprecated - I initially thought continuing to leave it out would be enough. What do you think?

benthorner:
Gotcha - I've added a warning to the readme and the code. Do you think that's enough?

PR #138, opened 2018-12-21
Show image credit on inline images

leenagupte:
Trello: 
* https://trello.com/c/VLAns9Hq
* https://trello.com/c/b0VF9sdO

## Motivation
Content publisher allows publishers to add an image credit for images they upload. Images in content publisher are being extended to allow publishers to add inline images, which will also allow an image credit.

Updates the govspeak gem so that image credits are rendered when users preview their content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,6 +473,7 @@ with options provided
         {
           alt_text: "Some alt text",
           caption: "An optional caption",
+          credit: "An optional credit",
           url: "http://example.com/lovely-landscape.jpg",
           id: "filename.png",
         }
@@ -485,7 +486,10 @@ will output a image section
 <figure class="image embedded">
   <div class="img">
     <img src="http://example.com/lovely-landscape.jpg" alt="Some alt text">
-    <figcaption>An optional caption</figcaption>
+    <figcaption>
+      <p>An optional caption</p>
+      <p>Image credit: An optional credit</p>
+    </figcaption>
   </div>
 </figure>
 ```
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -271,7 +271,7 @@ module Govspeak
       lines << %{<div class="img"><img src="#{encode(image.url)}" alt="#{encode(image.alt_text)}"></div>}
       lines << '<figcaption>' if image.figcaption?
       lines << %{<p>#{encode(image.caption)}</p>} if image.caption.present?
-      lines << %{<p>#{encode(image.credit)}</p>} if image.credit.present?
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -485,7 +486,10 @@ will output a image section
 <figure class="image embedded">
   <div class="img">
     <img src="http://example.com/lovely-landscape.jpg" alt="Some alt text">
-    <figcaption>An optional caption</figcaption>
+    <figcaption>
+      <p>An optional caption</p>
+      <p>An optional credit</p>
```
kevindew: Perhaps you want to prefix this with "Image credit:"

lib/govspeak.rb
```diff
@@ -269,11 +269,18 @@ def render_image(image)
       lines = []
       lines << %{<figure#{id_attr} class="image embedded">}
       lines << %{<div class="img"><img src="#{encode(image.url)}" alt="#{encode(image.alt_text)}"></div>}
-      lines << %{<figcaption>#{encode(image.caption)}</figcaption>} if image.caption
+      lines << '<figcaption>' if figcaption?(image)
```
kevindew: minor but since the subsequent lines inter join with the condition it might be clearer to just do all the figcaption stuff inside the condition. Which might mean you can drop the `figcaption?` method as I feel that's quite specific to be part of this `Govspeak::Document` class and more the domain of the presenter

lib/govspeak.rb
```diff
@@ -269,11 +269,18 @@ def render_image(image)
...
+      lines << %{<p>#{encode(image.caption)}</p>} if image.caption
+      lines << %{<p>#{encode(t('govspeak.image.figure.credit', credit: image.credit))}</p>} if image.credit
+      lines << '</figcaption>' if figcaption?(image)
       lines << '</figure>'
       lines.join
     end
 
+    def figcaption?(image)
+      image.caption.present? || image.credit.present?
```
kevindew: Any reason why we care about present? here but not so for `lines << %{<p>#{encode(image.caption)}</p>} if image.caption` ?

leenagupte: I normally use `present?` because I've been bitten a few times when evaluating more complex objects.  I'll make sure the lines are consistent.

lib/govspeak.rb
leenagupte: Fixed-up!

README.md
leenagupte: Updated and fixed-up.

lib/govspeak.rb
leenagupte: I've made this change. It's in a separate commit at the moment, but I will fix it up.

leenagupte: Updated and fixed up.

PR #141, opened 2019-03-07
Update to ruby 2.6.1

kevindew:
This reflects the most common Ruby version used on GOV.UK applications


kevindew:
Thanks @erino 👍

PR #142, opened 2019-03-07
Don't render [Image: <file>] that is within a paragraph

kevindew:
Trello: https://trello.com/c/MkWywL4g/590-support-inline-images

Doing this causes invalid HTML (it embeds elements such as figure, div
and p) and undesirable visual presentation - raw HTML is output.

Since this tag is not used in the wild just yet, it can be fixed before
users make use of it.

These same problems apply to the other Image embeds however I dare not
change those in case there are places where users are relying on the
invalid HTML is produced and it breaks documents.

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -29,6 +29,7 @@ module Govspeak
     Parser = Kramdown::Parser::KramdownWithAutomaticExternalLinks
     PARSER_CLASS_NAME = Parser.name.split("::").last
     UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.freeze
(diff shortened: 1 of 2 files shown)


lib/govspeak.rb
```diff
@@ -29,6 +29,7 @@ class Document
     Parser = Kramdown::Parser::KramdownWithAutomaticExternalLinks
     PARSER_CLASS_NAME = Parser.name.split("::").last
     UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.freeze
+    NEW_PARAGRAPH_LOOKBEHIND = %q{(?<=\A|\n\n|\r\n\r\n)}.freeze
```
emmabeynon: I noticed in the commit message that you mentioned a 'lookahead' but it's 'lookbehind' in the code so wanted to check what the correct term is?

kevindew: d'oh... thanks. It is indeed a lookbehind.

PR #144, opened 2019-03-13
Sanitize input by default

kevindew:
Trello: https://trello.com/c/SdMUjUmj/722-resolve-xss-issues-raised-by-pen-testers

This introduces a breaking change to govspeak where sanitization is performed by default on input and the means to turn this off is by passing a `sanitize: false` argument.

This replaces the previous method of sanitization which was to call `to_sanitized_html` on a govspeak document. This was a rather dubious approach as all testing of extensions was done against `to_html` so it was easy for some govspeak extensions to not work against that approach.

I had a ponder and I felt that since most apps use the [`valid?`](https://github.com/alphagov/govspeak/blob/34b2cd789b51ab9ce11cee5a0e8bfe3fac274da4/lib/govspeak.rb#L89) then the effects of turning sanitize on by default should be minimal. Any apps that aren't checking that would likely have security vulnerabilities that this resolves (as well as any pain points this introduces).

### DIFF
--- a/lib/templates/inline_attachment.html.erb
+++ b/lib/templates/inline_attachment.html.erb
@@ -1,6 +1,0 @@
-<span <% if attachment.id %>id="attachment_<%= attachment.id %>" <% end %>class="attachment-inline">
-  <%= attachment.link attachment.title, attachment.url %>
-  <% unless attachment.attachment_attributes.empty? %>
-    (<%= attachment.attachment_attributes %>)
-  <% end %>
-</span>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* BREAKING CHANGE: Input is sanitized by default, to use unsafe HTML initialize with a sanitize option of false
 * Allow sanitize option on remove invalid HTML from source input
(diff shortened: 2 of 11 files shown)


lib/templates/inline_attachment.html.erb
```diff
@@ -1,6 +0,0 @@
-<span <% if attachment.id %>id="attachment_<%= attachment.id %>" <% end %>class="attachment-inline">
```
benthorner: Should this be in another commit?

chao-xian:
I've only done a read through (which looks ok in that respect) but don't have enough context with our frontends to know what the impact is especially with a breaking change.

lib/templates/inline_attachment.html.erb
kevindew: Ta - sorted.

kevindew:
> You were right, this isn't pretty

I never said that 😿

> I think I'm missing some context here, though: why don't we just sanitise after all the processing is done, when only HTML remains?

That only works if all of our govspeak enhancements produce markup that meets our sanitising standards. Attachment is the rogue one that doesn't. 

Though, if anything, that is probably a sign that our sanitising is far too loose as allowing all those other govspeak extensions shows that we accept quite a variety of user input.

PR #150, opened 2019-05-02
 Use govuk_publishing_components to render attachment links

kevindew:
Trello: https://trello.com/c/zMYmFTPx/801-show-metadata-for-attachments

This is a draft PR until govuk_publishing_components is released and we can set that gem appropriately.

This sets the rendering of an AttachmentLink to be done via
govuk_publishing_components rather than done by govspeak logic.

It is moved to the post processing section so that there is no need to
be worried about new line characters breaking any markdown.

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -5,6 +5,7 @@ require 'htmlentities'
 require 'kramdown'
 require 'kramdown/parser/kramdown_with_automatic_external_links'
 require 'rinku'
+require 'govuk_publishing_components'
 require 'govspeak/header_extractor'
 require 'govspeak/structured_header_extractor'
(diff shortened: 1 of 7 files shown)


lib/govspeak.rb
```diff
@@ -5,6 +5,7 @@
 require 'kramdown'
 require 'kramdown/parser/kramdown_with_automatic_external_links'
 require 'rinku'
+require "govuk_publishing_components"
```
emmabeynon: Tiny annoying nitpick - this is in double quotes but the other requires have single quotes.

kevindew: Argh! Thanks for catching, I've fixed another one of those in the test files.

PR #154, opened 2019-05-22
Fix attachment link paragraphs

kevindew:
Trello: https://trello.com/c/zMYmFTPx/801-show-attachments-in-govspeak

This treats the custom element `<govspeak-embed-attachment-link />` as an inline element and thus is automatically wrapped in paragraph elements when used in govspeak.

### DIFF
--- a/test/govspeak_attachment_link_test.rb
+++ b/test/govspeak_attachment_link_test.rb
@@ -22,4 +22,19 @@ class GovspeakAttachmentLinkTest < Minitest::Test
     assert_match(/<span class="gem-c-attachment-link">/, rendered)
     assert_match(%r{href="http://example.com/attachment.pdf"}, rendered)
   end
+
+  test "renders an attachment link inside a paragraph" do
+    attachment = {
+      id: "attachment.pdf",
+      url: "http://example.com/attachment.pdf",
+      title: "Attachment Title",
+    }
+
+    rendered = render_govspeak("[AttachmentLink:attachment.pdf] test", [attachment])
+    root = Nokogiri::HTML.fragment(rendered).css(":root").first
+
+    assert(root.name, "p")
+    assert(root.css("p").size, 0)
+    assert_match(/Attachment Title\s*test/, root.text)
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fix wrapping `AttachmentLink:attachment-id` in a paragraph when used inline
+
 ## 6.1.0
 
 * Remove `[embed:attachments:content-id]` this isn't used by any apps and has never worked
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -3,7 +3,7 @@ require 'active_support/core_ext/array'
 require 'erb'
 require 'htmlentities'
 require 'kramdown'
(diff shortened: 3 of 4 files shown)


test/govspeak_attachment_link_test.rb
```diff
@@ -22,4 +22,15 @@ def render_govspeak(govspeak, attachments = [])
     assert_match(/<span class="gem-c-attachment-link">/, rendered)
     assert_match(%r{href="http://example.com/attachment.pdf"}, rendered)
   end
+
+  test "renders an attachment link inside a paragraph" do
```
benthorner: I'm wondering if we should have a couple more tests here:

   - The edge case of not inserting a paragraph when an attachment link is not found
   - A more specific test (possibly the same new one) to test the link and subsequent words are put in the same paragraph element

kevindew: We've got that first test you suggest: https://github.com/alphagov/govspeak/blob/d3b2914b1151a64d0b08a44b556b4b0fcf44b7be/test/govspeak_attachment_link_test.rb#L10-L12

Other test makes sense but it's a bit grim to do. I've had a go at it with nokogiri and pushed it up.

kevindew:
> Nice one @kevindew, although grizzly to be overriding the innards of Kramdown - perhaps we should just rewrite it ;-). I've suggested a couple more tests, but only if you think they'll add value.

Haha, would be nice to replace it with commonmarker but probably an utter nightmare.

> Totally out of scope, but it looks like opening up this particular part of Kramdown could enable us to avoid the [post processing we have to do for figures](https://github.com/alphagov/govspeak/blob/master/lib/govspeak/post_processor.rb#L25).

Yup, that would be good to kill off. That's only actually a problem because: https://github.com/alphagov/govspeak/blob/150fea5ef100420d84da7c40cbba7bada285a5ea/lib/govspeak.rb#L214 and https://github.com/alphagov/govspeak/blob/150fea5ef100420d84da7c40cbba7bada285a5ea/lib/govspeak.rb#L236 allow images inside a paragraph whereas https://github.com/alphagov/govspeak/blob/150fea5ef100420d84da7c40cbba7bada285a5ea/lib/govspeak.rb#L345 uses the new paragraph look behind. 

As all three of them use the same image presenter they could all use the same custom element and that block could be removed.

test/govspeak_attachment_link_test.rb
benthorner: Oh yes - sorry for missing that one.

test/govspeak_attachment_link_test.rb
```diff
@@ -22,4 +22,19 @@ def render_govspeak(govspeak, attachments = [])
...
+      url: "http://example.com/attachment.pdf",
+      title: "Attachment Title",
+    }
+
+    rendered = render_govspeak("[AttachmentLink:attachment.pdf] test", [attachment])
+    root = Nokogiri::HTML.fragment(rendered).css(":root").first
+
+    assert(root.name, "p")
+    assert(root.css("p").size, 0)
+    assert_match(/Attachment Title\s*test/, root.text)
```
benthorner: Oh dear, this isn't as simple as I thought it might be. I was hoping we could do some regex piece, but I can see how you got here. Up to you if we keep it though ;-).

kevindew: Yup regex was fraught with complications. I'll just merge this.

PR #156, opened 2019-05-30
Remove experimental status of Attachment extensions

kevindew:
Trello: https://trello.com/c/BQ5bBkMw/881-wrap-up-technical-loose-ends-on-attachments

These are now in production use by Content Publisher so their
experimental status has been removed.

As part of this the `[embed:attachments:inline:content-id]` extension
has been deprecated as this provides the same functionality as
`[AttachmentLink:attachment-id]`. At the time of writing this extension
is not directly used by any government publishers and is only used
behind the scenes in Specialist Publisher - A user enters a numeric
syntax and at the point of sending to Publishing API the govspeak is
converted to the inline syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Remove experimental status on `AttachementLink:attachment-id` and `Attachement:attachment-id`
+* Deprecate `embed:attachments:inline:content-id`
+
 ## 6.1.1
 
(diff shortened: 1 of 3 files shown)


kevindew:
> Looks good - is it important to show the HTML output in the README? It looks like it's the convention, but the actual rendered HTML might be more helpful ;-).

I doubt it it is that important - I'd agree screenshots would be nicer.

PR #159, opened 2019-06-26
Whitehall Govspeak upgrade

erino:
Whitehall has a number of documents created with previous versions of Govspeak which now fail validations. This is preventing Whitehall from using the latest version of Govspeak as publishers will no longer be able to updated published document.

I've captured some of the reasons why content is failing in 607aab05ce9e55be151f248ab6b90ee21728ccfa and later commits are to make the tests pass.

I'm not sure that this is the correct approach, which is why this PR is a draft is to start a discussion on the best way of fixing this problem.


kevindew:
Thanks for pushing this up Erin and great job with all the digging in.

I've had a bit of a play around myself and came to some conclusions.

For dealing with tbody we should change: https://github.com/alphagov/govspeak/blob/0e8aabd5fc090a67b7674bdb2992f1c33554b219/lib/govspeak/html_validator.rb#L23 to be `Nokogiri::HTML5.fragment(html).to_s`. The HTML5 rather than HTML will require nokogumbo, but that's ok since it's already required by sanitize. The `parse` => `fragment` should have no effect, just that we should have done that in the first place as parse spits out html, head and body elements that are superfluous. I believe this also resolves a lot of the odd quotes issues.

With the special characters I feel like we should just do a data migration find and replace on the govspeak_content body field to change them to a non-special counterpart. I guess that is \ufffc => \n and the rest to a space character?

Once I cut all of those out on my local machine I just have 7 that are still invalid:

```
[10] pry(main)> still_bad.map(&:id)
=> [1514055, 1514255, 3113168, 3124581, 3127439, 3128096, 3238660]
```

which can probably just be fixed manually.

test/html_validator_test.rb
```diff
@@ -102,43 +102,10 @@ class HtmlValidatorTest < Minitest::Test
...
   end
-
-  test "allow a table with a quotes spanning table elements" do
-    # The publication https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/964161
-    # with the HTML attachment "Upper Tribunal (Tax and Chancery) financial services hearings and register 2014 to date"
-    # has quotes spanning table elements
-    assert Govspeak::HtmlValidator.new("<table><tbody><tr><td>\"Hello</td><td>World\"</td></tr></tbody></table>").valid?
-  end
-
-  test "allow a govspeak table" do
```
benthorner: I guess this one also comes under 'valid HTML'.

PR #161, opened 2019-07-12
Add extension to allow table headers with scope

### LINKED ISSUE #110: Tables lacking row headers make them less accessible
Lots of tables on gov.uk are less accessible because they don't have row headers (i.e. a `th` as the first cell(s) in a row). If a cell cannot be understood without the context of the column header and/or row header, it is a barrier for screen reader users. Some tables need column headers, some need row headers and some need both, depending on context.

The main reason for missing row headers is that govspeak (or markdown/kramdown) doesn't support it.
(Whenever any Markdown flavour is asked to implement it, their usual response is "Use HTML!" even though an additional `th` syntax wouldn't make things much more complex and the majority of other plain text markup languages have a specific syntax for `th`s.)

A good syntax for a new implementation (either via a kramdown extension or a govspeak hack) I would like to propose is to add a `#` after the pipe. Using a `#` would be intuitive as it's already used for headings. For example:

```
|                 |# Second Column  |# Third Column          |
| :-------------- | :-------------: | ---------------------: |
|# First row      | Cell            | Very long data entry   |
|# Second row     | Cell            | Cell                   |
|# Third row      | Cell that spans across two columns      ||
```

(This example also includes a proposed syntax for merging cells, the second most needed feature to make govspeak tables more accessible.)

injms:
Tables need table headings to be accessible - screen readers and other assistive technologies use them to help explain the relationship between a heading and a cell's contents.

This pull request adds the ability set a cell to be a table heading (`th`), even when it's not inside the `thead` element.

In Govspeak the following markdown will have three `th` elements inside the top row because the `|---|` means that the top row should sit inside a `thead`:

```
|                | Second Column  | Third Column      |
| -------------- | -------------- | ----------------- |
| First row      | Cell           | Cell              |
```
will become: 

|                | Second Column  | Third Column      |
| -------------- | -------------- | ----------------- |
| First row      | Cell           | Cell              |

But there is currently no way to set the cell containing the text 'First row' to be a `th` element.

This pull request adds a tweak to the table syntax - by adding a `#` after the pipe we can set the cell to be a `th`. This is optional inside the table head, so the following two examples have the same outcome:

```
|                | Second Column  | Third Column      |
| -------------- | -------------- | ----------------- |
|# First row     | Cell           | Cell              |
```

```
|                |# Second Column |# Third Column     |
| -------------- | -------------- | ----------------- |
|# First row     | Cell           | Cell              |
```

Output:

```html
<table>
  <thead>
    <tr>
      <td></td>
      <th scope="col">Second Column</th>
      <th scope="col">Third Column</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th scope="row">First row</th>
      <td>Cell</td>
      <td>Cell</td>
    </tr>
  </tbody>
</table>
```

Fixes #110.

Ticket: https://trello.com/c/bRdKanmi/2-5-unsemantic-tables-23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 6.3.0
+## Unreleased
+
+ * Add table heading syntax that allows a table cell outside of `thead` to be marked as a table heading with a scope of row.
+
+ ## 6.3.0
 
 * Unicode characters forbidden in HTML are stripped from input
 * Validation is now more lenient for HTML input
--- a/lib/govspeak/post_processor.rb
+++ b/lib/govspeak/post_processor.rb
@@ -86,15 +86,16 @@ module Govspeak
     extension("Add table headers and row / column scopes") do |document|
       document.css("thead th").map do |el|
         el.content = el.content.gsub(/^# /, '')
-        el.content = el.content.gsub(/[[:space:]]/, '') if el.content.blank? # Removes a strange whitespace in the cell if the cell is already blank
-        el[:scope] = "col" if el.content.present?
+        el.content = el.content.gsub(/[[:space:]]/, '') if el.content.blank? # Removes a strange whitespace in the cell if the cell is already blank.
+        el.name = 'td' if el.content.blank? # This prevents a `th` with nothing inside it; a `td` is preferable.
+        el[:scope] = "col" if el.content.present? # `scope` shouldn't be used if there's nothing in the table heading.
       end
 
-      document.css(":not(thead) td").map do |el|
-        if el.content.match?(/^#.*$/)
-          el.name = 'th'
-          el[:scope] = 'row'
-          el.content = el.content.gsub(/^# /, '')
+      document.css(":not(thead) tr td:first-child").map do |el|
+        if el.content.match?(/^#($|\s.*$)/)
+          el.content = el.content.gsub(/^#($|\s)/, '') # Replace '# ' and '#', but not '#Word'.
+          el.name = 'th' if el.content.present? # This also prevents a `th` with nothing inside it; a `td` is preferable.
+          el[:scope] = 'row' if el.content.present? # `scope` shouldn't be used if there's nothing in the table heading.
         end
       end
     end
--- a/lib/govspeak/post_processor.rb
+++ b/lib/govspeak/post_processor.rb
@@ -83,6 +83,22 @@ module Govspeak
       end
     end
 
+    extension("Add table headers and row / column scopes") do |document|
+      document.css("thead th").map do |el|
+        el.content = el.content.gsub(/^# /, '')
+        el.content = el.content.gsub(/[[:space:]]/, '') if el.content.blank? # Removes a strange whitespace in the cell if the cell is already blank
+        el[:scope] = "col" if el.content.present?
+      end
+
+      document.css(":not(thead) td").map do |el|
+        if el.content.match?(/^#.*$/)
+          el.name = 'th'
+          el[:scope] = 'row'
+          el.content = el.content.gsub(/^# /, '')
+        end
+      end
+    end
+
     attr_reader :input, :govspeak_document
 
     def initialize(html, govspeak_document)
--- a/test/govspeak_table_with_headers_test.rb
+++ b/test/govspeak_table_with_headers_test.rb
@@ -1,0 +1,122 @@
+require 'test_helper'
+
+class GovspeakTableWithHeadersTest < Minitest::Test
+  def expected_outcome
+%{
+<table>
+  <thead>
+    <tr>
+      <th></th>
+      <th scope="col">Second Column</th>
+      <th scope="col">Third Column</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th scope="row">First row</th>
+      <td>Cell</td>
+      <td>Cell</td>
+    </tr>
+    <tr>
+      <th scope="row">Second row</th>
+      <td>Cell</td>
+      <td>Cell</td>
+    </tr>
+  </tbody>
+</table>
+}
+  end
+
+  def expected_outcome_with_hashes_in_cell_contents
+%{
+<table>
+  <thead>
+    <tr>
+      <th></th>
+      <th scope="col">Second Column</th>
+      <th scope="col">Third Column</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th scope="row">First row</th>
+      <td># Cell</td>
+      <td># Cell</td>
+    </tr>
+    <tr>
+      <th scope="row">Second row</th>
+      <td>Cell</td>
+      <td>Cell</td>
+    </tr>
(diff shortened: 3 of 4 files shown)


injms:
[(5) Unsemantic tables - {2,3}](https://trello.com/c/bRdKanmi/2-5-unsemantic-tables-23)

selfthinker:
It's a problem that the table header syntax is optional in the table head. When you have row headers, you usually don't want the first cell to be a header cell. In your code example the empty `<th></th>` would be a problem for screen reader users. Empty cells are generally a problem, but empty header cells are worse.

injms:
> It's a problem that the table header syntax is optional in the table head.

If the automatic table headers inside the table head were removed it would break the tables that are being used on the site at the moment. 

If we do want to make the `|#` syntax the _only_ way to define a table header:

* we first need to make them optional with backwards compatibility for old tables (which is what this pull request achieves) ◀️ 
* find all of the tables that are using the Kramdown table separator lines 🔎 
* change them to use the new syntax 🔀
* turn the old syntax off 🎆 

> When you have row headers, you usually don't want the first cell to be a header cell. In your code example the empty <th></th> would be a problem for screen reader users. Empty cells are generally a problem, but empty header cells are worse.

I've added a check that makes sure that an empty cell can't be a table header, so any blank cell will default to being a `td` without a scope attribute. As you say, a blank table cell isn't ideal - but better than a blank table header.

selfthinker:
Still having the table header syntax optional in the table head is fine with your changes. That's a good and probably even better solution. :+1:

Just for the record, as far as I know `scope` is not interpreted by any screen reader. It is following the spec correctly, so there is no harm in having it. But it makes more sense to have in more complex tables where simple `th`s are not enough.

selfthinker:
> Only allow table headings in the first column of a table
 > - A table heading should not be present more than once in a table row, except in the table head.

Why is that? Multiple table headers per row is perfectly valid and sometimes necessary. It shouldn't exactly be encouraged as we should always aim for the simplest possible table, but it's better and more accessible to allow it when it's needed than not have it and be confusing.

injms:
> > Only allow table headings in the first column of a table
> > 
> > * A table heading should not be present more than once in a table row, except in the table head.
> 
> Why is that? Multiple table headers per row is perfectly valid and sometimes necessary. It shouldn't exactly be encouraged as we should always aim for the simplest possible table, but it's better and more accessible to allow it when it's needed than not have it and be confusing.

I'd like us to bear in mind Markdown isn't supposed to be a 1:1 replacement for HTML and all of it's abilities:

> [Markdown is] "aimed at a sweet spot, between making it easy to use real HTML when you need it, and letting you just write plain text for anything where it’s sensible and obvious to do so." - [Dive into Markdown](https://daringfireball.net/2004/03/dive_into_markdown)

The sweet spot between the maximal and minimal examples of a table is what I think is achieved in this pull request - table headers can be used in the most common use cases using this Markdown-esque syntax. You're absolutely correct that multiple table headers per row are valid, useful, and accessible; and I'd suggest that that correct tool for making tables that require this complexity is HTML.

selfthinker:
But you can have multiple rows of column headers. Why not multiple columns of row headers? It's inconsistent to disallow one thing and not the other.

My understanding was that this feature was already in the code but that you have removed it?

injms:
> But you can have multiple rows of column headers. Why not multiple columns of row headers? It's inconsistent to disallow one thing and not the other.

I'd suggest that this isn't about consistency - it's about making the most common uses of tables easy to use in Govspeak. Tables that only require headers within the head and in the first column can use this syntax; more complex tables can use HTML.

accessiblewebuk:
There are situations where more than one row header is needed per row. For example a table where one of the headers isn't unique data. Simple example: 

'''


|   Street 1             | Street 2  | Traffic at intersection  |    
| -------------- | -------------| -------------- |
|Brick Lane    | Whitechapel Road    | lots           |
|Whitechapel Road     | Commercial Road    | gridlock           | 
|Brick Lane     | Commercial Road    | a bit         | 


For someone traversing down the Traffic at intersection column using a screenreader, they would need read out both street 1 and street 2 

Probably not a common use case, but I can't see any reason to restrict the number of row headers (or even column headers to one). Of course as @selfthinker points out, making data tables complex has its own issues from an accessibility/usability point of view, but we should try to leave the options open.

injms:
Thanks for your feedback. I agree that there are use cases for rows with multiple headers, but not that they should be in the Govspeak syntax. 

This covers the most common use cases where headers are needed, but it can't and it shouldn't try to cover every use case since there is already a well supported syntax that can make tables accessible: HTML. If we keep on adding extra pieces to Govspeak I think that we run the risk of making it as complicated as HTML, so instead of re-inventing the wheel we should use HTML.

As I've said before, Govspeak or Markdown isn't suppose to be a replacement for HTML since HTML can be used next to Markdown. Whitehall and Content Publisher use Govspeak and they do allow HTML in their content as part of the page editing process.

Since this pull request has been closed, I think that any further discussion should be raised as another issue so any work required can be properly prioritised.

selfthinker:
I don't understand why you say this would make things more complex and it would be "adding extra pieces to Govspeak". It is one single extra piece (the `|#` syntax) and it has already been added with this very PR.

I don't agree we "run the risk of making it as complicated as HTML". The table header syntax is not complicated. Markdown is one of the very few plain markup languages which doesn't support row headers. Even one of the smallest markup languages I know ([Creole](https://en.wikipedia.org/wiki/Creole_(markup))) has table header syntax. It's a very common element, not a special rarely used feature.

As far as I know, departments are not told anywhere if and how to use HTML in Whitehall/Publisher/etc?

Tobiogunsina:
I think I would agree with @selfthinker that allowing row header for multiple columns or just the 2nd column doesn't make it anymore complicated for us or for the publishers but may be beneficial for some use-cases. It's unclear to me why you think it does @injms 

Publishers should not need to add complicated tables to their document. Asking publishers to use HTML tables for things that are not complicated adds complexities for publishers and should be avoided. Especially as the content may need to be managed/edited by other publishers who may not have the same capability or understand how it's done. Avoiding html also means publishers are encouraged to keep things simple as opposed to making it complicated where it doesn’t need to be. 

Based on the experience of building content publisher, there is a huge variety in the capability of publishers and as a publishing principle, we aim to make publishing as intuitive as possible. Let's not take it for granted that lots of publishers will never need to deal with HTML otherwise. They shouldn't need to and neither should they have an additional separate workflow to do it. For example: the planned improvement for tables in Content publisher is to convert it to markdown for them. 

So considering the low level of complexity of what's needed, this decision should be based on the need and impact on publishers and GOV.UK users

**What will be helpful to make a decision is some more examples where this feature would have been useful and simplifying the table wouldn’t have helped.** 

I am also looking into why publishers are not told HTML is not allowed in cases where it's complicated enough. I imagine it’s for the same reasons as above but I’ll find out. Allowing HTML tables does seem like a separate issue though

PR #165, opened 2019-07-17
Bump to 6.3.0

kevindew:
* Unicode characters forbidden in HTML are stripped from input: https://github.com/alphagov/govspeak/pull/164
* Validation is now more lenient for HTML input: https://github.com/alphagov/govspeak/pull/159

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## 6.3.0
 
 * Unicode characters forbidden in HTML are stripped from input
 * Validation is now more lenient for HTML input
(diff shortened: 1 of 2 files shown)


benthorner:
I was thinking about this a bit more. Are we confident this won't have surprising consequences if we need to do a mass republishing?

kevindew:
> I was thinking about this a bit more. Are we confident this won't have surprising consequences if we need to do a mass republishing?

As discussed IRL. There is a chance it could cause some minor rendering problems to some pieces of content, but as the characters are not allowed in HTML it should be minimal. It affects such a low proportion of documents that it's unlikely to be a big concern and this same problem would be in effect if we stripped the chars from the db.

PR #167, opened 2019-08-23
Allow link data attributes

vanitabarrett:
## What
Allow any data attributes on links when sanitizing markdown

## Why
We have a use case for specifying content for a new landing page within a YML file. If this content contains a link which we want to track, at the moment this isn't possible as conversion of markdown to govspeak strips out the data attributes. In the case of ecommerce tracking, where we want to specify the linked-to page content id, it would be more difficult to add this logic into the presenter after the markdown has been converted.

**Note: I've gone with the approach of allowing any data attributes because tracking attributes can be varied and I'm not aware of any implications of allowing them all. Happy to switch to specific data attributes only if people prefer that approach**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Allow data attributes on links
+
 ## 6.4.0
 
  * Add table heading syntax that allows a table cell outside of `thead` to be marked as a table heading with a scope of row. (PR#161)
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@andysellick I'm guessing the default config is just as strict as possible without making it unusable, with the option to add in attributes as needed and we've just not had this use case until now 🤷‍♀

PR #173, opened 2019-12-13
Add rule to allow data attributes on divs

theseanything:
Whitehall uses data attributes on divs to provide accessibility links.

### DIFF
--- a/lib/govspeak/html_sanitizer.rb
+++ b/lib/govspeak/html_sanitizer.rb
@@ -55,6 +55,7 @@ class Govspeak::HtmlSanitizer
       attributes: {
         :all => Sanitize::Config::RELAXED[:attributes][:all] + %w[role aria-label],
         "a"  => Sanitize::Config::RELAXED[:attributes]["a"] + [:data],
+        "div" => [:data],
         "th"  => Sanitize::Config::RELAXED[:attributes]["th"] + %w[style],
         "td"  => Sanitize::Config::RELAXED[:attributes]["td"] + %w[style],
         "govspeak-embed-attachment" => %w[content-id],
--- a/test/html_sanitizer_test.rb
+++ b/test/html_sanitizer_test.rb
@@ -44,6 +44,15 @@ class HtmlSanitizerTest < Minitest::Test
     )
   end
 
+  test "allow data attributes on divs" do
+    html = "<div data-module='toggle' data-ecommerce-path='/' data-track-category='someDiv'>Test Div</div>"
+
+    assert_equal(
+      "<div data-module=\"toggle\" data-ecommerce-path=\"/\" data-track-category=\"someDiv\">Test Div</div>",
+      Govspeak::HtmlSanitizer.new(html).sanitize,
+    )
+  end
+
   test "allows images on whitelisted domains" do
     html = "<img src='http://allowed.com/image.jgp'>"
     sanitized_html = Govspeak::HtmlSanitizer.new(html, allowed_image_hosts: ["allowed.com"]).sanitize


barrucadu:
See https://github.com/alphagov/whitehall/blob/56006c6f6ba033fbe450ef91d46204499e62e337/app/views/documents/_attachment.html.erb#L75-L87:

```erb
    <% unless attachment.accessible? %>
      <div data-module="toggle" class="accessibility-warning" id="<%= help_block_id %>">
        <h2><%= t('attachment.accessibility.heading') %>
          <a class="toggler" href="#<%= help_block_toggle_id %>" data-controls="<%= help_block_toggle_id %>" data-expanded="false"><%= t('attachment.accessibility.request_a_different_format') %></a>
        </h2>
        <p id="<%= help_block_toggle_id %>" class="js-hidden">
          <%= t('attachment.accessibility.full_help_html',
                email: alternative_format_order_link(attachment, alternative_format_contact_email),
                title: attachment.title,
                references: attachment_references(attachment)) %>
        </p>
      </div>
    <% end %>
```

The `data-module="toggle"` is needed for the toggle link to work.  Without that, the `js-hidden` paragraph is hidden and there's no way to bring it back.  See this page for an example: https://www.gov.uk/government/statistical-data-sets/ofsted-parent-view-management-information

This is affecting all Whitehall pages which use whitehall's block attachment govspeak tag (`!@<number>`).  It doesn't affect publications because they don't use govspeak to render attachments: they render the attachment partial directly and concat the HTML.

barrucadu:
See https://github.com/alphagov/whitehall/blob/56006c6f6ba033fbe450ef91d46204499e62e337/app/views/documents/_attachment.html.erb#L75-L87:

```erb
    <% unless attachment.accessible? %>
      <div data-module="toggle" class="accessibility-warning" id="<%= help_block_id %>">
        <h2><%= t('attachment.accessibility.heading') %>
          <a class="toggler" href="#<%= help_block_toggle_id %>" data-controls="<%= help_block_toggle_id %>" data-expanded="false"><%= t('attachment.accessibility.request_a_different_format') %></a>
        </h2>
        <p id="<%= help_block_toggle_id %>" class="js-hidden">
          <%= t('attachment.accessibility.full_help_html',
                email: alternative_format_order_link(attachment, alternative_format_contact_email),
                title: attachment.title,
                references: attachment_references(attachment)) %>
        </p>
      </div>
    <% end %>
```

The `data-module="toggle"` is needed for the toggle link to work.  Without that, the `js-hidden` paragraph is hidden and there's no way to bring it back.  See this page for an example: https://www.gov.uk/government/statistical-data-sets/ofsted-parent-view-management-information

This is affecting all Whitehall pages which use whitehall's block attachment govspeak tag (`!@<number>`).  It doesn't affect publications because they don't use govspeak to render attachments: they render the attachment partial directly and concat the HTML.

kevindew:
@theseanything This isn't an ideal change to make as this allows users to potentially enter arbitrary HTML and data attributes can trigger javascript on the page. I don't think we've ever done an audit of what risks lie in data-modules on various apps but as a rule of thumb we don't want publishers to manually allow the embedding of javascript. 

The approach we have for things where javascript needs to be triggered for govspeak is to use custom HTML elements and then convert them into [HTML post-sanitisation](https://github.com/alphagov/govspeak/blob/master/lib/govspeak/post_processor.rb#L68-L84). This allows the output HTML to contain things that aren't allowed as input such as data attributes. Does that make sense?

I recognise that this might not be realistic for Whitehall if it's doing the attachment creation in Whitehall prior to govspeak conversion, is that the case? If that is so I wonder whether there is anything we can do to only whitelist very specific data attributes to not open up this hole. I hadn't noticed that we recently added the same for links.

PR #176, opened 2020-02-06
Use button component for buttons

andysellick:
- uses classes from govuk_publishing_components for buttons, calls component for start buttons (because start buttons include an SVG for the chevron)
- updates the HTML validator to allow SVG elements and their attributes
- this is part of deprecating the button class in static, which this relied on

Trello card: https://trello.com/c/d8D7JTJE/203-remove-design-patterns-buttons-from-static

Thanks to @chao-xian for pairing to fix tests!

### DIFF
--- a/README.md
+++ b/README.md
@@ -619,8 +619,7 @@ will output
 
 An accessible way to add button links into content, that can also allow cross domain tracking with [Google Analytics](https://support.google.com/analytics/answer/7372977?hl=en)
 
-This button component is [extended from static](http://govuk-static.herokuapp.com/component-guide/button) for [use in govspeak](http://govuk-static.herokuapp.com/component-guide/govspeak/button)
-Note: Ideally we'd use the original component directly but this currently isnt possible
+This button component uses the component from the [components gem](https://components.publishing.service.gov.uk/component-guide/button) for use in govspeak.
 
 You must use the [link](https://daringfireball.net/projects/markdown/syntax#link) syntax within the button tags.
 
@@ -635,7 +634,7 @@ To get the most basic button.
 which outputs
 
 ```html
-<a role="button" class="button" href="https://gov.uk/random">
+<a role="button" class="gem-c-button govuk-button" href="https://gov.uk/random">
   Continue
 </a>
 ```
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -619,8 +619,8 @@ will output
 
 An accessible way to add button links into content, that can also allow cross domain tracking with [Google Analytics](https://support.google.com/analytics/answer/7372977?hl=en)
 
-This button component is [extended from static](http://govuk-static.herokuapp.com/component-guide/button) for [use in govspeak](http://govuk-static.herokuapp.com/component-guide/govspeak/button)
-Note: Ideally we'd use the original component directly but this currently isnt possible
+This button component mimics the component from the [components gem](https://components.publishing.service.gov.uk/component-guide/button) for use in govspeak.
+Note: Ideally we'd use the original component directly but this currently isn't possible.
```
chao-xian: Can you add why it's not possible please? Maybe in the Description of the PR itself.

andysellick: Have reworked the code to make this line untrue, and removed it and modified the line above.

nickcolley:
:heart: :heart_eyes:

lib/govspeak/post_processor.rb
```diff
@@ -100,6 +100,24 @@ def self.extension(title, &block)
       end
     end
 
+    extension("use gem component for buttons") do |document|
+      document.css(".govuk-button").map do |el|
+        button_html = GovukPublishingComponents.render(
```
nickcolley: Love this

lib/govspeak.rb
```diff
@@ -161,10 +161,10 @@ def insert_strong_inside_p(body, parser = Govspeak::Document)
       {\/button} # match ending bracket
       (?:\r|\n|$) # non-capturing match to make sure end of line and linebreak
     }x) { |attributes, text, href|
-      button_classes = "button"
-      button_classes << " button-start" if attributes =~ /start/
+      button_classes = "govuk-button"
```
nickcolley: Could have considered using `GovukPublishingComponents.render` here instead of as a post process?

andysellick: Possibly, yes. To be honest I wasn't exactly sure how to do that, but that seems sensible. I might have another crack at it another time.

nickcolley: I think you'd replace line 176 with a call to the render function

nickcolley: e.g.

```ruby
GovukPublishingComponents.render(
  "govuk_publishing_components/components/button",
          text: text.strip,
          href: href.strip,
          start: (attributes =~ /start/),
          data_attributes: {
            module: 'cross-domain-tracking',
            "tracking-code": cross_domain_tracking.strip,
            "tracking-name": 'govspeakButtonTracker',
          },
        )
```

andysellick: Awesome, thanks!

nickcolley: Something I forgot in this example, that may be a bug in the existing code.
should only add those data attributes with `if cross_domain_tracking`

PR #179, opened 2020-06-17
Bump Sanitize gem

MuriloDalRi:


kevindew:
Also be useful to add an entry to the change log so gem users know why they're being asked to update.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.5.4
+
+* Update Sanitize to >= 5.2.1 < 6
```
kevindew, line 3:
```suggestion
## Unreleased

* Require Sanitize to be at least 5.2.1 due to https://nvd.nist.gov/vuln/detail/CVE-2020-4054
```

Unless we're releasing the gem in this PR (which we could do) we shouldn't change the version number in the changelog as that isn't yet determined.

PR #181, opened 2020-07-21
Extend default rake task with rubocop linting & fix lints

barrucadu:
[Trello card](https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,11 +8,11 @@ Bundler::GemHelper.install_tasks
 RuboCop::RakeTask.new
 
 desc "Run basic tests"
-Rake::TestTask.new("test") { |t|
+Rake::TestTask.new("test") do |t|
   t.libs << "test"
   t.pattern = "test/*_test.rb"
   t.verbose = true
   t.warning = true
-}
+end
 
 task default: %i[test rubocop]
--- a/Rakefile
+++ b/Rakefile
@@ -1,8 +1,11 @@
 require "rake"
 require "rake/testtask"
+require "rubocop/rake_task"
 require "bundler"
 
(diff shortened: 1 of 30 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Fix linting issues
```
kevindew, line 1: I don't think this really needs to go in the changelog as it won't affect consumers of the gem (we hope).

Rakefile
```diff
@@ -5,11 +5,16 @@ require "bundler"
...
   t.verbose = true
   t.warning = true
-}
+end
 
-task default: [:test]
+desc "Lint Ruby"
+task :lint do
+  sh "bundle exec rubocop --format clang"
+end
```
kevindew, line 18: We can use the built in rubocop task rather than defining our own. 

So this could have a require earlier of `require 'rubocop/rake_task'` and initialise `RuboCop::RakeTask.new` in this file then the default task can be `task default: %i[test rubocop]` without needing a task defined.

Rakefile
```diff
@@ -1,15 +1,18 @@
 require "rake"
 require "rake/testtask"
+require "rubocop/rake_task"
 require "bundler"
 
 Bundler::GemHelper.install_tasks
 
+RuboCop::RakeTask.new(:lint)
```
kevindew, line 8: There probably isn't any need to deviate from the default of rubocop here. Often our `lint` tasks do all the types of linting of an app and rubocop can be specific for just ruby linting.

PR #187, opened 2020-08-19
Prevent links in table headers being stripped

maxgds:
The substitution that strips # from table headers is effectively
taking a copy of the text but not the child html which means that
links and other declared-safe html is lost. We know the # that
indicates a header will always appear at the start and therefore be
in the first child so this targets the substitution to the content of
the first child instead of the whole content. Tests added to confirm
expected behaviour persists.


test/govspeak_table_with_headers_test.rb
```diff
@@ -131,6 +131,39 @@ def expected_outcome_for_table_with_blank_table_headers
...
+    <tr>
+      <th scope="row"><a rel="external" href="http://www.bbc.co.uk">Whole header is link</a>
+</th>
+      <td>Cell</td>
+      <td>Cell</td>
+    </tr>
+  </tbody>
+</table>
+)
+  end
```
owenatgov, line 165: This might be overkill so please guide me, but for this test and the one below using govspeak syntax should there be some assertions to ensure that the link is rendering? I'm not familiar with govspeak's testing stack so as I say, this might be too much and I might just be being daft.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Allow links in table headers with a scope of row
```
andysellick, line 3: I think this should match the PR title and include the PR number, as per previous entries.

test/govspeak_table_with_headers_test.rb
maxgds: I think you're missing what's happening. This is the expected output and the govspeak syntax is the input. The actual test is further down and compares the two.

lib/govspeak/post_processor.rb
```diff
@@ -94,7 +94,9 @@ def self.extension(title, &block)
       document.css(":not(thead) tr td:first-child").map do |el|
```
andysellick, line 94: The PR description is talking about this problem being to do with table headers, but this selector seems to be specifically not table headers, can you explain?

lib/govspeak/post_processor.rb
```diff
@@ -94,7 +94,9 @@ def self.extension(title, &block)
...
         next unless el.content.match?(/^#($|\s.*$)/)
 
-        el.content = el.content.gsub(/^#($|\s)/, "") # Replace '# ' and '#', but not '#Word'.
+        # Replace '# ' and '#', but not '#Word'.
+        # This runs on the first child of the element to preserve any links
+        el.children.first.content = el.children.first.content.gsub(/^#($|\s)/, "")
```
andysellick, line 99: I'm not sure I understand this so I may be asking a nonsense question. This previous ran on the content of the cell, but is now running on the first child of the cell. Does it still do what it was supposed to do if the cell has no children?

lib/govspeak/post_processor.rb
maxgds: This is targetting table headers that exist outside of the table head. They appear in the left hand column and have a scope of "row". The ones in the table head are dealt with elsewhere (Kramdown, I think) and we had to add this extension to allow row headers.

lib/govspeak/post_processor.rb
maxgds: If the cell has content, that content is the child (children.first) and it works. 
If the cell has content that includes a link it has multiple children.
If the cell has no content it is picked up later and the TH is turned into a TD.

CHANGELOG.md
maxgds: Will fix.

test/govspeak_table_with_headers_test.rb
owenatgov: 🙈 Oops. Comment withdrawn.

lib/govspeak/post_processor.rb
andysellick: Ah, I think the confusion is due to the overlap when talking about "table headers" - the difference between `thead` and `th` elements. What's the element being targeted here? This line appears to refer to the first `td` in a `tr` that is not inside a `thead` - is that correct?

lib/govspeak/post_processor.rb
andysellick: Okay, so even text content e.g. `This is the header` is counted as a child and processed correctly? (the `children.first` you referred to)

lib/govspeak/post_processor.rb
maxgds: Yes, that's right, <th> outside the <thead> in the first column with a scope of row.

lib/govspeak/post_processor.rb
maxgds: Yes, see the test I added and note the input and output data.

maxgds: `<td>I am <a href="">a link</a> in a sentence</td>` - this breaks down as 3 children - text child, link child, text child. But the content of each can be targeted, and confusingly (what caused the bug) targeting content on the cell collects the content of each.

andysellick: Thanks, that's clarified things.

andysellick (approved):

PR #192, opened 2020-10-26
Customise footnote markup for accessibility 

danacotoran:
The default footnote "backlink" only has an arrow as content. 
Screen readers only say "right arrow curving left" when they get to this element.
This is a problem in terms as accessibility as it is unclear to the users what those links do. A better solution is to use text to indicate what the link does, which is what this is aiming to achieve.

This commit changes the footnote markers from being just numbers to **footnote 1,2,3...i**. Similarly, it replaces the "back link" which is not very helpful to the user, with **"Go to where this is referenced"**, which describes precisely what the link should do.

-----

Example of a page where footnotes are being used abundantly: https://www.gov.uk/government/publications/budget-2016-documents/budget-2016

-----

### Backlink before
Screenshot 2020-10-26 at 19 21 30

### Backlink after

Screenshot 2020-10-28 at 17 22 30

### Footnote before
Screenshot 2020-10-26 at 19 22 48

### Footnote after

Screenshot 2020-10-28 at 17 23 32

https://trello.com/c/SJyttlVo/459-footnotes-footnote-links-and-backlinks-unclear

### DIFF
--- a/test/govspeak_footnote_test.rb
+++ b/test/govspeak_footnote_test.rb
@@ -1,0 +1,31 @@
+require "test_helper"
+require "govspeak_test_helper"
+
+class GovspeakFootnoteTest < Minitest::Test
+  include GovspeakTestHelper
+
+  test_given_govspeak "
+    Footnotes can be added[^1].
+
+    [^1]: And then later defined.
+
+    Footnotes can be added too[^2].
+
+    [^2]: And then later defined too." do
+    assert_html_output '
+      <p>Footnotes can be added<sup id="fnref:1" role="doc-noteref"><a href="#fn:1" class="footnote">[footnote 1]</a></sup>.</p>
+
+      <p>Footnotes can be added too<sup id="fnref:2" role="doc-noteref"><a href="#fn:2" class="footnote">[footnote 2]</a></sup>.</p>
+
+      <div class="footnotes" role="doc-endnotes">
+        <ol>
+          <li id="fn:1" role="doc-endnote">
+            <p>And then later defined. <a href="#fnref:1" class="reversefootnote" role="doc-backlink">[go to where this is referenced]</a></p>
+          </li>
+          <li id="fn:2" role="doc-endnote">
+            <p>And then later defined too. <a href="#fnref:2" class="reversefootnote" role="doc-backlink">[go to where this is referenced]</a></p>
+          </li>
+        </ol>
+      </div>'
+  end
+end
--- a/lib/govspeak/post_processor.rb
+++ b/lib/govspeak/post_processor.rb
@@ -120,6 +120,16 @@ module Govspeak
       end
     end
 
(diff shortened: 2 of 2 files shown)


chris-gds:
image

✅   As expected

chris-gds (approved):

test/govspeak_footnote_test.rb
```diff
@@ -0,0 +1,31 @@
+require "test_helper"
+require "govspeak_test_helper"
+
+class GovspeakFootnoteTest < Minitest::Test
+  include GovspeakTestHelper
+
+  test_given_govspeak "
+    Footnotes can be added[^1].
+
+    [^1]: And then later defined.
```
chris-gds, line 10: I've one question about this `[^bignote]` [use ](https://www.markdownguide.org/extended-syntax#footnotes) and if this is something that we need to factor in?

Eg
`Here's a simple footnote,[^1] and here's a longer one.[^bignote]`

This doesn't render the same way but then it also doesn't render on govspeak master so prob out of scope.

danacotoran: Ooh interesting. I don't think we support or encourage `[^bignote]` notation? 
This is the suggested formatting in whitehall publisher.
So if I understand correctly, seems that we do support multiline footnotes but they should always be numbered. 


Screenshot 2020-11-04 at 14 49 16

stevenjmesser:
I've raised a related issue in the Kramdown repo, a Ruby Markdown converter: https://github.com/gettalong/kramdown/issues/692

danacotoran:
> I've raised a related issue in the Kramdown repo, a Ruby Markdown converter: [gettalong/kramdown#692](https://github.com/gettalong/kramdown/issues/692)

Thanks for this @stevenjmesser. There's very little information on accessible footnotes out there, and most of the examples we've tested on the web were not accessible; it would be great if accessible footnotes could become more widespread. 

It should be worth mentioning that I've just opened a follow up PR (#198) as I found some issues with the original implementation while testing.

stevenjmesser:
No problem, @danacotoran, it could really help by making thousands of blogs more accessible. 

If we wanted to, I think we could potentially raise a PR on the Kramdown repo. Aside from several tests, I think I've identified the two lines that need updating:

1. https://github.com/gettalong/kramdown/blob/1b8fd33c3120bfc6e5164b449e2c2fc9c9306fde/lib/kramdown/options.rb#L544
2. https://github.com/gettalong/kramdown/blob/1b8fd33c3120bfc6e5164b449e2c2fc9c9306fde/lib/kramdown/options.rb#L573

What do you think?

test/govspeak_footnote_test.rb
chris-gds: > I don't think we support or encourage [^bignote] notation?
This is the suggested formatting in whitehall publisher.

Makes sense.

PR #195, opened 2020-11-02
Remove reference to advisory callout

keithiopia:
We haven't recommend the use of advisory 'highlights' for a long time – this readme might be the only place they're still referenced. They also have an issue with colour constrast.
Also removed the 'answer' highlight. We don't use those on mainstream anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ creates an example box
 
 ## Highlights
 
-### Advisory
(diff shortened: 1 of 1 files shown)


kevindew:
@keithiopia it feels like we should also be removing the code for these if we're removing the documentation. We can probably do a query across content to check nothing current still uses them and if so delete the code and the docs at the same time.

kevindew (approved):

PR #198, opened 2020-11-05
Adjust footnote markup to accommodate multiple references

danacotoran:
This is a follow up on the original work done for footnote accessibility here #192 

# The Problem
The original PR had replaced the footnote arrow "back link" with the words "go to where this is referenced"  like so:

Screenshot 2020-11-05 at 18 13 03

This was done with the aim of making those links clearer to both sighted and non-sighted users.

However as we were testing the original PR, we discovered an edge case where the solution that was implemented does not work.
The edge case appears when footnotes are referenced multiple times in the body of the publication, and therefore contain multiple numbered "back links", as can be seen [here](https://www.gov.uk/government/publications/spring-budget-2017-overview-of-tax-legislation-and-rates-ootlar/annex-a-rates-and-allowances#fn:13 ).

### Multiple numbered back arrow links
Screenshot 2020-11-05 at 18 31 43

With the changes introduced in #192, a footnote that is referenced multiple times would look like this, which is obviously an issue: 

Screenshot 2020-11-05 at 18 17 47

# The Solution 
Publishers should ideally not be using footnotes as they are not great on the web.
In the meanwhile, we're implementing a middle ground solution. This reverts the footnote "back link" to using the arrow, with the adjustment that the back link will have an `aria-label` of **"go to where this is referenced"** or **"go to where this is referenced 2,3,4... i"** when there's more than one reference. This should help AT users navigate this area, while keeping it fairly terse and understandable for sighted users. 

https://trello.com/c/SJyttlVo

### DIFF
--- a/lib/govspeak/post_processor.rb
+++ b/lib/govspeak/post_processor.rb
@@ -126,7 +126,8 @@ module Govspeak
         el.content = "[footnote #{footnote_number}]"
       end
       document.css("[role='doc-backlink']").map do |el|
-        el.content = "[go to where this is referenced]"
+        backlink_number = " " + el.css("sup")[0].content if el.css("sup")[0].present?
+        el["aria-label"] = "go to where this is referenced#{backlink_number}"
       end
     end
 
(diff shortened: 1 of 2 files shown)


injms (approved):
Good solution 👍

PR #202, opened 2020-12-30
Fix issue rendering $CTA followed by $C block

theseanything:
There was a issue were the $C in the first $CTA tag was being matched with the final $C tag. This cause the block of content to be rendered as a contact instead of a call to action. This changes the priority of the regexs evaluated so $CTA tags identified and parsed before $C preventing the mis-match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.5.11
+
+* Fix issue rendering $CTA blocks before $C (PR#202)
+
 ## 6.5.10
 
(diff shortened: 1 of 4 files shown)


kevindew (approved):
Nice catch. Gosh it's all so fragile in here 😬 .

You've got a typo in your commit message by the way, I think it might be worth also leaving a comment to say that $CTA has to be defined before $C for specificity - It feels like the sort of nuance you'd want documented beyond a commit message.

theseanything:
Thanks @kevindew! Added a comment and fixed the spelling in the commit.

PR #203, opened 2021-01-13
Allow optional elements to skip sanitisation

edwardkerry:
This change allows an optional array of `relaxed_sanitization_elements` to be passed to a new Document, and these will be excluded from HTML sanitization.

There is a use-case where Whitehall is submitting a `<details>` element which is being stripped by sanitization.

[Trello](https://trello.com/c/uQpgMD83/2289-5-accessible-format-request-show-hide-not-working-on-detailed-guide-pages)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.6.0
+
+* Allow passed elements to be relaxed from sanitization [#203](https://github.com/alphagov/govspeak/pull/203)
+
 ## 6.5.11
 
 * Fix issue rendering $CTA blocks before $C (PR#202)
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -53,6 +53,7 @@ module Govspeak
       @source = source ? source.dup : ""
 
       @images = options.delete(:images) || []
+      @allowed_elements = options.delete(:allowed_elements) || []
       @attachments = Array.wrap(options.delete(:attachments))
       @links = Array.wrap(options.delete(:links))
       @contacts = Array.wrap(options.delete(:contacts))
@@ -66,7 +67,7 @@ module Govspeak
     def to_html
       @to_html ||= begin
                      html = if @options[:sanitize]
-                              HtmlSanitizer.new(kramdown_doc.to_html).sanitize
+                              HtmlSanitizer.new(kramdown_doc.to_html).sanitize(allowed_elements: @allowed_elements)
                             else
                               kramdown_doc.to_html
                             end
--- a/lib/govspeak/html_sanitizer.rb
+++ b/lib/govspeak/html_sanitizer.rb
@@ -40,18 +40,19 @@ class Govspeak::HtmlSanitizer
     @allowed_image_hosts = options[:allowed_image_hosts]
   end
 
-  def sanitize
+  def sanitize(allowed_elements: [])
     transformers = [TableCellTextAlignWhitelister.new]
     if @allowed_image_hosts && @allowed_image_hosts.any?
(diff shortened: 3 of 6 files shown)


1pretz1 (approved):
LGTM 🍏

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.5.12
```
kevindew, line 1: This is probably 6.6.0 as this is adding a new feature.

lib/govspeak.rb
```diff
@@ -53,6 +53,7 @@ def initialize(source, options = {})
       @source = source ? source.dup : ""
 
       @images = options.delete(:images) || []
+      @relaxed_sanitization_elements = options.delete(:relaxed_sanitization_elements) || []
```
kevindew, line 56: This seems quite a long argument that's a bit tricky to understand I somewhat wonder if there is a nicer name for it.  It seems a little bit coupled to "relaxed" of sanitize which seems like something a user of this gem wouldn't need to know about.

I'll have a quick ponder for suggestions.

lib/govspeak/html_sanitizer.rb
```diff
@@ -40,12 +40,16 @@ def initialize(dirty_html, options = {})
...
-  def sanitize
+  def sanitize(relaxed_elements: [])
     transformers = [TableCellTextAlignWhitelister.new]
     if @allowed_image_hosts && @allowed_image_hosts.any?
       transformers << ImageSourceWhitelister.new(@allowed_image_hosts)
     end
-    Sanitize.clean(@dirty_html, Sanitize::Config.merge(sanitize_config, transformers: transformers))
+
+    config = sanitize_config
+    relaxed_elements.each { |el| config[:elements].add(el) }
```
kevindew, line 50: I feel like it'd be a bit easier to follow of these elements were added as part of `sanitize_config` given we already have our own elements added in there: https://github.com/alphagov/govspeak/blob/5e514e4d7c2f5b37a793098b9fc6a532d7921ee7/lib/govspeak/html_sanitizer.rb#L58

Perhaps this should be an argument to `sanitize_config`

test/govspeak_test.rb
```diff
@@ -666,6 +666,11 @@ class GovspeakTest < Minitest::Test
     assert_equal "<script>doGoodThings();</script>", document.to_html.strip
   end
 
+  test "it can exclude stipulated elements from sanitization" do
+    document = Govspeak::Document.new("<uncommon-element>some content</uncommon-element>", relaxed_sanitization_elements: %w[uncommon-element])
+    assert_equal "<uncommon-element>some content</uncommon-element>", document.to_html.strip
+  end
```
kevindew, line 672: Nice one 👍

kevindew (commented):
Awesome stuff Ed, a few comments that I hopefully get in before you merge.

lib/govspeak.rb
kevindew: perhaps just simply `allowed_elements` ?

thomasleese: I had a similar thought when looking at this code, I agree with `allowed_elements`.

CHANGELOG.md
edwardkerry: Thanks, I was in two minds about if this counted as a feature - but agree it is!

lib/govspeak.rb
edwardkerry: Thanks both, renamed 👍

thomasleese (approved):
This looks good to me 👍🏻

PR #206, opened 2021-02-08
Update heading & docs

chris-gds:
## What?

Update `govspeak` contact information to remove bold

## Why?

Contact information should not be in bold. This is confusing as it looks like a heading but is not  [(WCAG 1.3.1)](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html).  Also at present it is a `h3` which wasn't always accurate in terms of a `h2` always being before it in the context of the page

## Visual Changes
Before (left) After (right)
image

## Anything else?

`govspeak` is used by the gem [there is a dependant PR here.](https://github.com/alphagov/govuk_publishing_components/pull/1909)

### DIFF
--- a/lib/templates/contact.html.erb
+++ b/lib/templates/contact.html.erb
@@ -4,8 +4,8 @@
 %>
 <div id="contact_<%= contact.content_id %>" class="<%= ['contact', extra_class].flatten.join(' ') %>">
   <div class="content">
-    <h3><%= contact.title %></h3>
     <div class="vcard contact-inner">
+    <p><%= contact.title %></p>
     <% contact.post_addresses.each do |address| %>
       <%= Govspeak::HCardPresenter.new(address).render %>
     <% end %>
--- a/README.md
+++ b/README.md
@@ -598,8 +598,8 @@ will output
 ```html
 <div id="contact_123" class="contact">
   <div class="content">
-    <h3>Government Digital Service</h3>
     <div class="vcard contact-inner">
+      <p>Government Digital Service</p>
       <div class="email-url-number">
         <p class="email">
           <span class="type">Email</span>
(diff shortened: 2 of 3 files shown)


lib/templates/contact.html.erb
```diff
@@ -4,8 +4,8 @@
 %>
 <div id="contact_<%= contact.content_id %>" class="<%= ['contact', extra_class].flatten.join(' ') %>">
   <div class="content">
-    <h3><%= contact.title %></h3>
     <div class="vcard contact-inner">
+    <p><%= contact.title %></p>
```
maxgds, line 8: I'm intrigued that this has moved position as well as losing the header designation - not saying it's wrong, but what's the reasoning behind it?

chris-gds: Ah so, my thinking was that previously it was header, so it was placed _outside_ to indicate that this is different to the `contact-inner` content. However now, this is no longer the case, it's all one-of-the-same semantically so why is it sitting outside?  Also, if there is anything that is applied to `vcard contact-inner` that I can't see this in is update (maybe styling elsewhere?) this update can receive them.  However equally,  happy to update if preferred.

maxgds: It's probably fine either way, but being cautious it might be best to keep the position the same for your last reason! If there is something out there targetting the inner then maybe it's something we wouldn't want this to inherit? But honestly, it probably makes no difference, I'm happy to approve this as is, just wanted to understand if there was reaoning.

maxgds (approved):

lib/templates/contact.html.erb
chris-gds: It's true, both could be true, I think the second part to this ticket might shed a bit more light either way on this one. Hopefully it leans towards solving an issue than causing one, thanks for approving.

Issue #101, opened 2016-10-12
Phone number links in contacts / text?


edent:
## Use Case

As a user, I want to be able to click on a phone number from my mobile browser and have my phone call the number.

At the moment, phone numbers are rendered in plain text. Depending on the browser, they *may* be clickable.

Additionally, the user may not know that they need to add an international dialling code (`+44` for UK numbers).

## Suggestions

* Repurpose the email syntax, so `<020 7946 0123>` is automatically converted to `<a href="tel:+442079460123>020 7946 0123</a>`
* Validate the regular Markdown of `[020 7946 0123](tel:+442079460123)`
* Use a magic token like `$P` to indicate a **P**hone Number.

## Additional
Perhaps add some metadata / microformat to help with indexing / search?

## References

* http://www.rfc-editor.org/rfc/rfc3966.txt

keithiopia:
Has there been any evidence to suggest users need this or would understand why a phone number is a link on desktop/laptops?

edent:
Only that I tried to contact the DVLA via GOVUK and it was very annoying that I had to copy & paste the phone number.

Tried in Chrome, Opera, and Firefox for Android.

I'm not sure how best to go about testing whether regular users find that a barrier to contacting people.

As for clicking on desktop - a `tel` link tends to open Facetime on a Mac, and other dialers on other OSes.  I'm not sure if that's different from an auto-linking email address if someone doesn't have an email program installed,

keithiopia:
I could flag it with the user research people if you like? 
I'm sure it could be incorporated into some other GOV.UK benchmarking.

edent:
👍

edent:
As an example of a site which uses the tel schema https://peoplefinder.cabinetoffice.gov.uk/people/terence-eden

edent:
Just as a note, over a third of visitors use a mobile device https://www.gov.uk/performance/site-activity/device-type

edent:
As an alternative - we have a [mandated standard for contact exchange](https://www.gov.uk/government/publications/open-standards-for-government/exchange-of-contact-information).

Would it be sensible to have [vCard](https://en.wikipedia.org/wiki/VCard)s on all "contact us" pages?

keithiopia:
That's an interesting idea - but would devices get updated when departments/helplines change their number or address?

edent:
No - but that's no different from today. I assume that departments which change numbers / emails / postal addresses maintain the old one for a suitable period of time.

[...]

PR #207, opened 2021-02-09
Update to link telephone numbers

chris-gds:
## What?

Update so telephone numbers are clickable for users. 

## Why?

While working on [this PR](https://github.com/alphagov/govuk_publishing_components/pull/1909) I thought this would improve the UX and potentially the a11y for users.

This update allows a user to click a telephone number and have the relevant app responsible for calling activated.

Automatic format detection exists for many mobile device browsers, this will convert mobile numbers into clickable numbers automatically. On desktop there are other behaviours.  At present a user will have to copy-paste or write down a number in order to call whilst viewing on desktop. 

There are different behaviours on desktop: on macOS this might result in FaceTime Audio or tethering being activated. On Windows this might activate VoIP (eg Skype), on Chromebook there are other behaviours.

## Anything else?

- [Noticed there is further conversations about this happening here](https://github.com/alphagov/govspeak/issues/101)
 
- Draft PR for comments / technical reference
- This would need investigation with users / a support matrix of behaviours
- There are other considerations and testing required (eg international numbers)
- [The Design System currently notes _not_ to do this, however acknowledges further research needed to validate](https://design-system.service.gov.uk/patterns/telephone-numbers/)

Further reading ([Link 1](https://css-tricks.com/the-current-state-of-telephone-links/), [Link 2](https://developers.google.com/web/fundamentals/native-hardware/click-to-call))

---

## Related issues and PRs

https://github.com/alphagov/govuk-design-system/issues/2022

https://github.com/alphagov/govuk-design-system/pull/2070

### DIFF
--- a/lib/templates/contact.html.erb
+++ b/lib/templates/contact.html.erb
@@ -26,7 +26,7 @@
       <% contact.phone_numbers.each do |number| %>
         <p class="tel">
           <span class="type"><%= number[:title] %></span>
-          <%= number[:number] %>
+          <a href="tel:<%= number[:number] %>"><%= number[:number] %></a>
         </p>
       <% end %>
       </div>


chris-gds:
Accessibility Community meeting noted that this is being revisited and there are benefits to user

kevindew:
Closing since this seems to be stale. Can be reopened if this catches someones attention again.

PR #216, opened 2021-09-06
Fix footnotes in legislative lists

deborahchua:
## What
This fixes an issue with footnotes within legislative lists in Whitehall.
Trello card: https://trello.com/c/yfeqz1no/2561-footnotes-dont-work-within-legislativelist-on-manuals-publisher-5

## Why
This has been a longstanding issue and is heavily used in legislation papers. 
It means that publishers cannot use footnotes within certain kinds of content, and so cannot always follow the design guidelines. Having markdown that's reliable and effective is helpful for selling our accessibility approach (avoiding PDFs) to policy colleagues.

## Visual Changes
Before (footnotes in a legislative list not converted into html):
Screenshot 2021-09-06 at 12 39 21

After (footnotes in a legislative list now showing):
Screenshot 2021-09-06 at 12 40 29

Before (footnote definitions missing):
Screenshot 2021-09-06 at 12 44 13

After (footnote definitions now showing):
Screenshot 2021-09-06 at 12 48 05

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -63,6 +63,7 @@ module Govspeak
                    syntax_highlighter: nil }.merge(options)
       @options[:entity_output] = :symbolic
       @footnote_definition_html = nil
+      @acronyms = []
     end
 
     def to_html
@@ -135,6 +136,7 @@ module Govspeak
     def legislative_list_footnote_definitions(source)
       is_legislative_list = source.scan(/\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m).size.positive?
       footnotes = source.scan(/\[\^(\d)\]:(.*)/)
+      @acronyms = source.scan(/(?<=\*)\[(.*)\]:(.*)/)
 
       if is_legislative_list && footnotes.size.positive?
         list_items = footnotes.map do |footnote|
@@ -158,6 +160,10 @@ module Govspeak
             </ol>
           </div>
         HTML_CONTAINER
+      end
+
+      unless @footnote_definition_html.nil? && @acronyms.size.positive?
+        add_acronym_alt_text(@footnote_definition_html)
       end
     end
 
@@ -348,6 +354,8 @@ module Govspeak
 
             doc.sub!(/(\[\^#{footnote}\])/, html)
           end
+
+          add_acronym_alt_text(doc) if @acronyms.size.positive?
         end
       end
     end
@@ -438,6 +446,12 @@ module Govspeak
 
     def encode(text)
       HTMLEntities.new.encode(text)
+    end
+
+    def add_acronym_alt_text(html)
+      @acronyms.each do |acronym|
+        html.gsub!(acronym[0], "<abbr title=\"#{acronym[1].strip}\">#{acronym[0]}</abbr>")
+      end
     end
   end
 end
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -140,11 +140,12 @@ module Govspeak
         list_items = footnotes.map do |footnote|
           number = footnote[0]
           text = footnote[1].strip
+          footnote_definition = Govspeak::Document.new(text).to_html[/(?<=<p>).*(?=<\/p>)/]
 
           <<~HTML_SNIPPET
             <li id="fn:#{number}" role="doc-endnote">
               <p>
-                #{text}<a href="#fnref:#{number}" class="reversefootnote" role="doc-backlink" aria-label="go to where this is referenced">↩</a>
+                #{footnote_definition}<a href="#fnref:#{number}" class="reversefootnote" role="doc-backlink" aria-label="go to where this is referenced">↩</a>
               </p>
             </li>
           HTML_SNIPPET
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -62,6 +62,7 @@ module Govspeak
                    sanitize: true,
                    syntax_highlighter: nil }.merge(options)
       @options[:entity_output] = :symbolic
+      @footnote_definition_html = nil
     end
 
     def to_html
@@ -71,6 +72,16 @@ module Govspeak
                else
                  kramdown_doc.to_html
                end
+
+        unless @footnote_definition_html.nil?
+          regex = /<div class="footnotes".*[<\/div>]/m
+
+          if html.scan(regex).empty?
+            html << @footnote_definition_html
+          else
+            html.gsub!(regex, @footnote_definition_html)
+          end
+        end
 
         Govspeak::PostProcessor.process(html, self)
       end
@@ -110,12 +121,43 @@ module Govspeak
     def preprocess(source)
       source = Govspeak::BlockquoteExtraQuoteRemover.remove(source)
       source = remove_forbidden_characters(source)
+
+      legislative_list_footnote_definitions(source)
+
       self.class.extensions.each do |_, regexp, block|
         source.gsub!(regexp) do
           instance_exec(*Regexp.last_match.captures, &block)
         end
       end
       source
+    end
+
+    def legislative_list_footnote_definitions(source)
+      is_legislative_list = source.scan(/\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m).size.positive?
+      footnotes = source.scan(/\[\^(\d)\]:(.*)/)
+
+      if is_legislative_list && footnotes.size.positive?
+        list_items = footnotes.map do |footnote|
+          number = footnote[0]
+          text = footnote[1].strip
+
+          <<~HTML_SNIPPET
+            <li id="fn:#{number}" role="doc-endnote">
+              <p>
+                #{text}<a href="#fnref:#{number}" class="reversefootnote" role="doc-backlink" aria-label="go to where this is referenced">↩</a>
+              </p>
+            </li>
+          HTML_SNIPPET
+        end
+
+        @footnote_definition_html = <<~HTML_CONTAINER
+          <div class="footnotes" role="doc-endnotes">
+            <ol>
+              #{list_items.join.strip}
+            </ol>
+          </div>
+        HTML_CONTAINER
+      end
     end
 
     def remove_forbidden_characters(source)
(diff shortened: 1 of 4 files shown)


lib/govspeak.rb
```diff
@@ -62,6 +62,7 @@ def initialize(source, options = {})
                    sanitize: true,
                    syntax_highlighter: nil }.merge(options)
       @options[:entity_output] = :symbolic
+      @footnote_definition_html = ""
```
ChrisBAshton, line 65: We appear to be using an empty string to replicate nil-ish behaviour here.
It would probably be better to set this to `nil` initially, and change line 76 to `unless @footnote_definition_html.nil?`.

lib/govspeak.rb
```diff
@@ -295,6 +329,19 @@ def render_image(image)
           doc.gsub!("<ul>", "<ol>")
           doc.gsub!("</ul>", "</ol>")
           doc.sub!("<ol>", '<ol class="legislative-list">')
+
+          footnotes = body.scan(/\[\^(\d)\]/)
```
ChrisBAshton, line 342: This will only catch the first 9 footnotes. I doubt we'd ever need more than that, but just in case, perhaps worth changing to:

```suggestion
          footnotes = body.scan(/\[\^(\d+)\]/)
```

You probably want to avoid the `footnotes[counter][0]` syntax later on too, by flattening the structure here, so that you could just use `footnotes[counter]`:

```suggestion
          footnotes = body.scan(/\[\^(\d+)\]/).flatten
```

lib/govspeak.rb
```diff
@@ -295,6 +329,19 @@ def render_image(image)
...
+          counter = 0
+
+          while footnotes.size.positive? && counter < footnotes.size
```
ChrisBAshton, line 345: It would be more idiomatic to use something like `footnotes.each_with_index do |footnote, i|` and remove the `counter` variable altogether.

lib/govspeak.rb
```diff
@@ -118,6 +132,26 @@ def preprocess(source)
       source
     end
 
+    def legislative_list_footnote_definitions(source)
```
ChrisBAshton, line 136: I've got some comments in the `remove_forbidden_characters` method which equally apply to this one.

But that does make me wonder, can we reuse any of the code, rather than duplicate the logic in both places? It feels like we could DRY things up by changing the internal API to something like:

```ruby
list_items = each_footnote(footnotes).map do | footnote, index|
  <<HTML_SNIPPET
    <li id="fn:#{index}" role="doc-endnote">
      <p>
        #{footnote}<a href="#fnref:#{index}" class="reversefootnote" role="doc-backlink" aria-label="go to where this is referenced">↩</a>
      </p>
    </li>
  HTML_SNIPPET
end


@footnote_definition_html = <<HTML_CONTAINER
  <div class="footnotes" role="doc-endnotes">
    <ol>
      #{list_items}
    </ol>
  </div>
HTML_CONTAINER
```

lib/govspeak.rb
```diff
@@ -141,9 +141,14 @@ def legislative_list_footnote_definitions(source)
         @footnote_definition_html = "<div class=\"footnotes\" role=\"doc-endnotes\">\n<ol>"
 
         while counter < footnotes.size
+          footnote_definition = Govspeak::Document.new(footnotes[counter][1].strip).to_html[/(?<=<p>).*(?=<\/p>)/]
```
ChrisBAshton, line 146: Nice workaround 👍 my only concern is the regex inside the `to_html` here. I guess this is to add `<p>` tags to the footnotes, which replicates how the 'normal' footnotes work, but if ever there is a change to the HTML that footnotes should be nested within, we'd have to remember to update this one too.

Is there a common presenter somewhere that we can delegate this logic to rather than having it inline here?

lib/govspeak.rb
```diff
@@ -155,6 +157,8 @@ def legislative_list_footnote_definitions(source)
 
         @footnote_definition_html << "</ol></div>"
       end
+
+      add_acronym_alt_text(@footnote_definition_html) if @acronyms.size.positive?
```
ChrisBAshton, line 161: Nice one 👏

ChrisBAshton (commented):
Well done for working through this fiddly issue! I've got some code style comments below, but otherwise this looks basically ready to go. However, there is something I'd like to check first. From the first commit: 

> We currently don't have the ability to render footnotes in Legislative
> lists. This workaround substitutes the footnote anchor links (e.g. [^1]
> for it's equivalent html.

Do we know _why_ we don't have the ability to render footnotes in Legistlative lists? Is there just a method call we need to invoke somewhere? It feels awfully cumbersome to have to recreate our own footnotes/acronyms markdown parsing logic if it already exists and works elsewhere. Apologies if you've already looked into this, I'm viewing this with fresh eyes and haven't looked too closely at the coal face.

My only other question was whether we're duplicating [any of this code](https://github.com/alphagov/govspeak/blob/8a4a1341109c393ebcb68d2daaf635489da48553/lib/govspeak/post_processor.rb#L123-L132) - it particularly feels as though the "go to where this is referenced" bit should be encapsulated in one place rather than repeated here.

lib/govspeak.rb
deborahchua: Ah so this was something handy I discovered which uses using positive lookbehind and positive lookahead in the regex - `/(?<=<p>).*(?=<\/p>)/` finds all the characters between a `<p>` and `</p>` tag, but does not include the tags.
From what I understand they will always be wrapped in `<p>` tags if the spacing is correct in the markdown.

ChrisBAshton: Ah ok - so the `Govspeak::Document.new` takes the footnote markdown, we call `.to_html` on it to convert it from markdown to HTML, but we use the regex bit to strip out the `<p>` tags because they'll be added in automatically later on? Is that right?

deborahchua: Yep that's right!

deborahchua:
Thanks @ChrisBAshton! Addressing your comments below:

> > We currently don't have the ability to render footnotes in Legislative
> > lists. This workaround substitutes the footnote anchor links (e.g. [^1]
> > for it's equivalent html.
> 
> Do we know _why_ we don't have the ability to render footnotes in Legistlative lists? Is there just a method call we need to invoke somewhere? It feels awfully cumbersome to have to recreate our own footnotes/acronyms markdown parsing logic if it already exists and works elsewhere. Apologies if you've already looked into this, I'm viewing this with fresh eyes and haven't looked too closely at the coal face.

So this was the part I spent a great deal of time trying to figure out a better workaround. Leglislative lists was a new markdown tag introduced a while back so that editors can have custom numbering for ordered lists and different indentation levels (see [this commit](https://github.com/alphagov/govspeak/commit/489fa7099f602c740bef5ccdeedc15c52033efd1) which explains in detail) The approach used means a [sub document](https://github.com/alphagov/govspeak/blob/main/lib/govspeak.rb#L294) is created and the footnotes only exists inside that document, with some hacks to the Kramdown to make it work. 
Going by the previous comments and my understanding of the problem, this icky workaround seemed like the best option without having to create our own custom markdown parser for lists.

> My only other question was whether we're duplicating [any of this code](https://github.com/alphagov/govspeak/blob/8a4a1341109c393ebcb68d2daaf635489da48553/lib/govspeak/post_processor.rb#L123-L132) - it particularly feels as though the "go to where this is referenced" bit should be encapsulated in one place rather than repeated here.

The html for footnotes doesn't exist at this stage if the document has footnotes in legislative lists, which is why it needs to be manually processed at an [earlier point](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R126)

lib/govspeak.rb
deborahchua: Thanks @ChrisBAshton , I've updated this now - see [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R65) and [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R77)

lib/govspeak.rb
deborahchua: Thanks @ChrisBAshton I hadn't noticed that! I also forgot about the `flatten` method and looks much neater now.
 I've updated it now [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R347-R354)

lib/govspeak.rb
deborahchua: Thanks @ChrisBAshton , was clearly looking at this for too long and didn't see this improvement when refactoring! 
It doesn't require the `index` as the footnote numbers can vary and exist in multiple Legislative lists, so the values are simply stored in `footnotes` - I've now updated it [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R349-R355)

lib/govspeak.rb
deborahchua: Thanks @ChrisBAshton , this is definitely a much more elegant way of dealing with the HTML! I've updated it [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R146-R161)

lib/govspeak.rb
deborahchua: This has been updated to accommodate the use of `nil` in `@footnote_definition_html`, see [here](https://github.com/alphagov/govspeak/pull/216/files#diff-acb6db6cd67285bd5380398c9e51ced8940ed7c4110706d038cbb243310e4055R164-R166)

lib/govspeak.rb
```diff
@@ -118,6 +132,33 @@ def preprocess(source)
       source
     end
 
+    def legislative_list_footnote_definitions(source)
+      regex = /\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m
```
ChrisBAshton, line 137: We only use the regex in the conditional on line 139 - we could probably take the same approach as you've taken for `footnotes` on the next line, and do the scanning here:

```suggestion
      is_legislative_list = source.scan(/\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m).size.positive?
```

...then use this on line 139 as `if is_legislative_list && footnotes.size.positive?`

lib/govspeak.rb
```diff
@@ -118,6 +132,33 @@ def preprocess(source)
...
+      footnotes = source.scan(/\[\^(\d)\]:(.*)/)
+
+      if source.scan(regex).size.positive? && footnotes.size.positive?
+        list_items = footnotes.map do |footnote|
+          number = footnote[0]
+
```
ChrisBAshton, line 145: I'd be tempted to add a `text = footnote[1].strip` here, so that the way we're dealing with the `footnote` is consistent for both the text and number.

ChrisBAshton (approved):
Thanks for the explanation, and for making the changes, @deborahchua! This looks good to go 👍  I've got a couple of nitpick comments below, but they're non-blocking.

PR #218, opened 2021-09-15
Fix footnote regex in Legislative lists

deborahchua:
This should now capture all footnote numbers rather than limiting it
to 9 and was missed in https://github.com/alphagov/govspeak/pull/216


ChrisBAshton (approved):
Nice one 👍

Hjordaan2002 (approved):

PR #220, opened 2021-09-30
Remove Nokogumbo dependency

kevindew:
## What

This removes the Nokogumbo dependency both directly from this gem, and indirectly by updating Sanitize

## Why

Nokogumbo is no longer needed as an additional dependency see: https://github.com/sparklemotion/nokogiri/issues/2205. It's presence leaves a warning message each time it is required of `NOTE: nokogumbo: Using Nokogiri::HTML5 provided by
Nokogiri. See https://github.com/sparklemotion/nokogiri/issues/2205 for more information.`

## Visual Changes

None


leenagupte (approved):
👍

PR #224, opened 2021-11-19
Add draggable false to button markup and validation

owenatgov:
## What
Adds `draggable="false"` to the govspeak button extension output.

## Why
This is in response to a change to button markup in the components gem (https://github.com/alphagov/govuk_publishing_components/pull/2448) which wasn't reflected in govspeak. This caused the govspeak validator to assume all buttons failed as the in-app markup was including `draggable="false"` whilst sanitised govspeak wasn't.

[Zendesk ticket which prompted this](https://govuk.zendesk.com/agent/tickets/4779605)

No visual changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Add draggable false to button markup and validation [#224](https://github.com/alphagov/govspeak/pull/224)
+
 ## 6.7.5
 
 * Fix footnotes in call-to-action boxes [222](https://github.com/alphagov/govspeak/pull/222)
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -224,7 +224,7 @@ module Govspeak
       text = text.strip
       href = href.strip
 
-      %(\n<a role="button" class="#{button_classes}" href="#{href}" #{data_attribute}>#{text}</a>\n)
+      %(\n<a role="button" draggable="false" class="#{button_classes}" href="#{href}" #{data_attribute}>#{text}</a>\n)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.7.6
```
karlbaker02, line 1: Is the convention that we make an `Unreleased` commit first and then make another commit to bump the version?

lib/govspeak.rb
```diff
@@ -224,7 +224,7 @@ def insert_strong_inside_p(body, parser = Govspeak::Document)
       text = text.strip
       href = href.strip
 
-      %(\n<a role="button" class="#{button_classes}" href="#{href}" #{data_attribute}>#{text}</a>\n)
+      %(\n<a role="button" draggable="false" class="#{button_classes}" href="#{href}" #{data_attribute}>#{text}</a>\n)
```
karlbaker02, line 227: We still need `draggable="false"` here right?

karlbaker02 (commented):
Looks good, a few small comments and questions

CHANGELOG.md
owenatgov: I was thinking this. I've been looking back at some old PRs and I reckon that convention makes sense. I'll update accordingly.

lib/govspeak.rb
owenatgov: After a very quick test, we indeed do.

karlbaker02 (approved):
Looks great, thanks Owen! 👍

PR #225, opened 2021-11-19
Bump release to 6.7.6

owenatgov:
* Add draggable false to button markup and validation [#224](https://github.com/alphagov/govspeak/pull/224)


karlbaker02 (approved):
Nice one 👍

PR #233, opened 2022-02-14
Update to Ruby 2.7.5 and Rails 7

ollietreend:
This PR adds support for using this gem with Rails 7. It also changes the repository to use Ruby 2.7.5, since Ruby 2.6 is soon reaching End of Life (EOL).

The CI pipeline has been updated to test this gem against Ruby versions `2.7`, `3.0` and `3.1`.

Trello: https://trello.com/c/j0uMp0Y9


govspeak.gemspec
```diff
@@ -29,7 +29,7 @@ library for use in the UK Government Single Domain project'
   s.executables   = s.files.grep(%r{^bin/}) { |f| File.basename(f) }
   s.require_paths = %w[lib]
 
-  s.add_dependency "actionview", ">= 5.0", "< 7"
+  s.add_dependency "actionview", ">= 5.0"
```
kevindew, line 32: We may as well bump this to `>= 6` since Rails 5 doesn't run on Ruby 2.7. It does seem reasonable to be optimistic with the versioning here.

govspeak.gemspec
```diff
@@ -42,7 +42,6 @@ library for use in the UK Government Single Domain project'
   s.add_development_dependency "minitest", "~> 5.14"
   s.add_development_dependency "pry-byebug"
   s.add_development_dependency "rake"
-  s.add_development_dependency "rubocop-govuk", "~> 4.2.0"
+  s.add_development_dependency "rubocop-govuk", "~> 4"
```
kevindew, line 45: I think you might want to actually set this as a fixed dependency of "4.2.0" (without the `~>`) as it can be annoying when the linting of this differs depending where you're running it. 

For example, if you install locally and have rubocop-govuk 4.0 you may pass the linting locally but then have it fail in CI as it's installed 4.2.

lib/govspeak.rb
```diff
@@ -1,3 +1,4 @@
+require "active_support"
```
kevindew, line 1: I couldn't understand from the commit why this change was needed.

kevindew (commented):
Looks great - just a few minor snags

lib/govspeak.rb
ollietreend: @kevindew Fair shout – I've just broken this out into its own commit to document the need: 8dbdcf8aa5487672ad70f0977325574f3ce995ce

.github/workflows/ci.yml
```diff
@@ -6,7 +6,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.6, 2.7]
+        # Due to https://github.com/actions/runner/issues/849, we have to use quotes for '3.0'
```
kevindew, line 9: Ooh that's interesting

govspeak.gemspec
```diff
@@ -42,7 +42,6 @@ library for use in the UK Government Single Domain project'
   s.add_development_dependency "minitest", "~> 5.14"
   s.add_development_dependency "pry-byebug"
   s.add_development_dependency "rake"
-  s.add_development_dependency "rubocop-govuk", "~> 4.2.0"
+  s.add_development_dependency "rubocop-govuk", "4.3.0"
```
kevindew, line 45: Sneaky upgrade 👍

govspeak.gemspec
```diff
@@ -29,7 +29,7 @@ library for use in the UK Government Single Domain project'
   s.executables   = s.files.grep(%r{^bin/}) { |f| File.basename(f) }
   s.require_paths = %w[lib]
 
-  s.add_dependency "actionview", ">= 5.0", "< 7"
+  s.add_dependency "actionview", ">= 6.0"
```
kevindew, line 32: Pedantic nitpick, do we need the ".0"?

lib/govspeak.rb
kevindew: Great thanks. Does that mean we don't need the next 2 requires? presumably active_support pulls them in?

kevindew (commented):
This looks great to me, just 2 super minor things

lib/govspeak.rb
ollietreend: As I understand it, no:

```ruby
require "active_support"
```

just pulls in the bare minimum dependencies required to cherry-pick the features you want.

Whereas:

```ruby
require "active_support/all"
```
is what you'd use if you wanted to pull in the whole kitchen sink.

At least that's how I'm interpreting it based on the docs – see: [1.1.1 Cherry-picking a Definition](https://guides.rubyonrails.org/active_support_core_extensions.html#cherry-picking-a-definition) vs [1.1.4 Loading All Active Support](https://guides.rubyonrails.org/active_support_core_extensions.html#loading-all-active-support)

---

Somewhat concerningly, the test suite seems to pass even when I remove/comment out all the `active_support` imports entirely 😳 – i.e. this passes:

```ruby
# require "active_support"
# require "active_support/core_ext/hash"
# require "active_support/core_ext/array"
```

But it seems cherry-picking without the initial `active_support` include is where things break.

kevindew: Oh cool TIL - yes that makes sense.

As for your concern, as I understand govuk_publishing_components pulls in various activesupport bits, so they probably just get satisfied a few lines later. All a bit clunky compared to JS/Python

kevindew (approved):
Looks good to me, if you're still working now save merging until tomorrow in case we see any problems with the gem release

benjamineskola (approved):

PR #234, opened 2022-02-16
Use GitHub Actions for continuous integration

ollietreend:
## What

This PR configures the repository to use GitHub Actions for continuous integration instead of Jenkins.

It's mostly a copy & paste of the [CI configuration for `rubocop-govuk`](https://github.com/alphagov/rubocop-govuk/blob/main/.github/workflows/ci.yml) which also uses GitHub Actions to test and publish to RubyGems.

## Why

To adopt [GOV.UK RFC-123](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-123-github-actions-ci.md).

## Visual Changes

None. This is a dev workflow change, so there are no changes to the gem itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,45 @@
+name: Continuous integration
+on: [push, pull_request]
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.6, 2.7]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true
+    - run: bundle exec rake
+  
+  release:
+    needs: test
+    runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/main' }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
+
+      # Update to >= v3.0.5
+      # https://gitlab.com/gitlab-org/gitlab-qa/-/issues/431
+    - run: gem update --system
+
+    - env:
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -1,7 +0,0 @@
-#!/usr/bin/env groovy
-
-library("govuk")
-
-node {
-  govuk.buildProject()
-}
```
ollietreend, line 7: Does anything else need to be done to remove this project from Jenkins?

benjamineskola (approved):

Jenkinsfile
benjamineskola: It's possible that it's referenced in alphagov/govuk-puppet.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
+name: Continuous integration
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
```
kevindew, line 9: It'd be good to test a matrix of Ruby versions here - I'd imagine 2.7, 3.0 and 3.1 are right

ollietreend: Definitely 👍🏻 I intend to do that, but right now it doesn't seem to run in `3.0` due to some out of date gems.

I was intending for this PR to simply port the CI over to GitHub Actions but keep it on Ruby 2.6.

The PR #233 would follow, in which I'll implement a matrix test (which should then hopefully pass) against 2.7, 3.0 and 3.1.

Let me know if you think there's a better approach though!

kevindew: Sure, that makes sense. Do you want to set this up as a matrix of 2.6 and 2.7 first? It'd be good to have a matrix test in on the first instance incase it reveals any problems

PR #239, opened 2022-03-29
Fix footnote numbering

baisa:
## What
Fix bug with footnote numbering in a legislative list

## Why
To prevent from showing empty additional footnotes.

[Trello card](https://trello.com/c/R23P4vYE/315-investigate-and-fix-bug-with-footnote-numbering-in-a-legislative-list)

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -138,7 +138,7 @@ module Govspeak
     def footnote_definitions(source)
       is_legislative_list = source.scan(/\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m).size.positive?
       is_cta = source.scan(/\$CTA.*?\[\^\d\]*.*?\$CTA/m).size.positive?
-      footnotes = source.scan(/\[\^(\d+)\]:(.*)/)
+      footnotes = source.scan(/^\s*\[\^(\d+)\]:(.*)/)
       @acronyms = source.scan(/(?<=\*)\[(.*)\]:(.*)/)
       if (is_legislative_list || is_cta) && footnotes.size.positive?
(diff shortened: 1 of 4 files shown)


lib/govspeak.rb
```diff
@@ -138,7 +138,7 @@ def preprocess(source)
     def footnote_definitions(source)
       is_legislative_list = source.scan(/\$LegislativeList.*?\[\^\d\]*.*?\$EndLegislativeList/m).size.positive?
       is_cta = source.scan(/\$CTA.*?\[\^\d\]*.*?\$CTA/m).size.positive?
-      footnotes = source.scan(/\[\^(\d+)\]:(.*)/)
+      footnotes = source.scan(/\[\^(\d+)\]:(.*)/).select { |footnote| footnote[-1].present? }
```
kevindew, line 141: I'm a bit confused by how this fixes the problem - the code looks rather cryptic. Would you be able to explain more in the commit message? I'm also not sure I fully understand what the cause of the problem is.

Are you using footnote[-1] as a means to access the last item in an arrays match? if so would `footnote.last` be less cryptic?

I'm also wondering will this bug not still exist if they had any extra characters on a legislative list line. For example:

```
    $LegislativeList
    1.  some text[^1]: extra
    $EndLegislativeList
    [^1]: footnote text
```

or even just some spaces?

Perhaps we should only be matching a footnote definition if there is only whitespace before one? for example `/^\s*\[\^(\d+)\]:(.*)/`

baisa: Correcting regex is a better solution, thank you @kevindew I have added this additional test to make sure it works as expected.

ollietreend (approved):

PR #245, opened 2022-06-15
Use shared workflow to publish gem

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This makes use of the shared workflow that has been added to
alphagov/govuk-infrastructure which allows publishing rubygems.


BeckaL (approved):
LGTM 👍

PR #246, opened 2022-06-21
Bump kramdown >=2.3.0 ->  >=2.3.1

willmcb:
Bump kramdown by minor version to patch security issue:

https://github.com/advisories/GHSA-52p9-v744-mwjj

## What
dependency minor version bump

## Why
fixes security issue with gem dependency

## Visual Changes
n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.8.3
+
+* Require Kramdown minimum version of 2.3.1 to avoid CVE-2021-28834 [#246](https://github.com/alphagov/govspeak/pull/246)
+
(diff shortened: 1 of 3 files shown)


Davidslv:
Just to clarify, This is used at MOJ, we don't want to fork govspeak. Thanks for having a look

kevindew:
Thank you very much for opening this ⭐ 

As it's a requirement to be ">= 2.3.0" you shouldn't need to do any forking to avoid the vulnerability as apps using it will accept a later kramdown (looks like ours are all using Kramdown 2.4.0).

kevindew (approved):

PR #249, opened 2022-07-22
Remove `PriorityList` from Govspeak

MartinJJones:
## What
Remove `PriorityList`, including related tests and documentation.

## Why
Priority List is no longer required as the Primary Links JS module used to render the list on the frontend it to be removed from govuk-publishing-components.

## Visual Changes

| Before | After |
| - | - |
| before | after |

Trello: https://trello.com/c/nezjGrzF/1388-investigate-and-potentially-fix-primary-links-module


KludgeKML (approved):
LGTM

BeckaL (commented):
Looks good, and always nice to remove code! Small nitpick, could you squash the last two commits that bump the version into one? Would be nice if the bump to 6.9 doesn't appear in the git history since we decided to go with a major bump in the following commit (which absolutely seems the correct call to me 😄 )

MartinJJones:
Thanks @BeckaL, good point!, I've squashed the last 2 commits as recommended 👍

BeckaL (approved):
Nice, LGTM!

BeckaL (approved):

kevindew:
I'm a bit worried that prior to merging this we'll need to adapt content of anything that uses this and update guidance. I wrote more on [the components issue](https://github.com/alphagov/govuk_publishing_components/issues/2364)

MartinJJones:
Latest update rebased main to resolve merge conflicts

PR #268, opened 2023-01-27
Replace inline style attributes in td/th elements with classes

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

We aspire to disallow inline style elements on GOV.UK as these present a XSS security risk [1]. However this change was scuppered [2] because Govspeak has one scenario (text alignment in Kramdown) which uses inline styles [3].

The approach I've taken to resolve this is to use the post-processor as a way to replace the style elements on td and th elements with class parameters. This change won't be safe to be merged until govuk_publishing_components is updated to apply the styles of the class.

This handles a couple of edge cases, that may not be that necessary. It can cope if the style attribute is given multiple text-align rules and will use the last one (as CSS would). It also can handle the scenario where this already a class on the td/th.

Once the GOV.UK stack migrates existing content to use it can remove the leniency that allows style elements in the sanitizer rules. I've left a comment as a reminder that this could be removed (hopefully I'll remember).

[1]: https://github.com/alphagov/govuk_app_config/blob/13083f8f8563c4703c14fe8175ca35646ec64442/lib/govuk_app_config/govuk_content_security_policy.rb#L64-L67
[2]: https://github.com/alphagov/govuk_app_config/pull/279#issuecomment-1398721551
[3]: https://github.com/gettalong/kramdown/blob/0b0a9e072f9a76e59fe2bbafdf343118fb27c3fa/test/testcases/block/14_table/header.html#L37-L56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Replace inline style attributes in td/th elements with classes [#268](https://github.com/alphagov/govspeak/pull/268)
+
 ## 7.0.2
 
 * Fix for abbreviations nested in button. [#267](https://github.com/alphagov/govspeak/pull/267)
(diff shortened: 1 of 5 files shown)


BeckaL (approved):
Looks good to me, thanks for useful links and commit messages.

PR #274, opened 2023-05-18
Make image and attachment embedding syntax more consistent

ollietreend:
## What

This PR makes changes to the `[Image:]` and `[Attachment:]` embed syntaxes so they behave more consistently with the older `!!n` (image) and `!@n` (attachment) syntaxes.

In particular, `[Image:]` and `[Attachment:]` only worked when preceded by two newline characters (`\n\n`). Whereas `!!n` and `!@n` only needed to be preceded by one newline character (`\n`).

### Before

For example, this would work when embedding with the `!!n` syntax:

```
A paragraph of text.
!!1
Another paragraph of text.
```

But this **would not work** with the `[Image:]` syntax:

```
A paragraph of text.
[Image: example.jpg]
Another paragraph of text.
```

### After

Both examples above work as expected.

They both produce HTML output like:

```html
<p>A paragraph of text.</p>
<figure><!-- image is embedded here --></figure>
<p>Another paragraph of text.</p>
```

## Why

We've recently introduced [support for `[Image:]` syntax in Whitehall][1], and will soon be adding support for `[Attachment:]` too. We quickly received feedback from users that the new image embedding syntax wasn't working – and it turns out it was because they were used to embedding images using `!!n` syntax, which only needed one newline character.

To resolve the inconsistency, we've decided to bring the syntaxes in line by making the newer `[Image:]` and `[Attachment:]` syntaxes more lenient as users were expecting.

[1]: https://github.com/alphagov/whitehall/pull/7589

Trello: https://trello.com/c/FxDTL9Ic/1272-update-image-syntax-so-it-behaves-similar-to-images-added-in-whitehall-previously-no-line-break-needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 7.1.1
+
+* Make image and attachment embedding syntax more consistent [#274](https://github.com/alphagov/govspeak/pull/274)
+
 ## 7.1.0
 
 * Drop support for Ruby 2.7 [#272](https://github.com/alphagov/govspeak/pull/272)
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -424,7 +424,7 @@ module Govspeak
       renderer.render(contact: ContactPresenter.new(contact))
     end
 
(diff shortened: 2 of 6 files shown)


BeckaL (approved):
LGTM, thanks!

PR #277, opened 2023-06-29
Wrap the text of start buttons in a span

CristinaRO:
When using an inline element (such as `<abbr>`, `<address>` or `<telephone>`) in a link with the `govuk-button--start` class applied to it, there is no spacing between the inline element and the text of the link.

The reason that this is occurring is the `display: inline-flex;` rule in the `govuk-button--start` class.

The suggested solution was to wrap the text of `start` buttons in a `<span>` tag:
https://github.com/alphagov/govuk-frontend/issues/3406#issuecomment-1563139563

This is the solution I'm implementing here.

I've done a quick test in integration, covering only a start button with an `abbr` in it.

## Before
Screenshot 2023-06-29 at 16 32 13

## After
Screenshot 2023-06-29 at 16 35 25

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -225,6 +225,7 @@ module Govspeak
         data_attribute << " data-tracking-name='govspeakButtonTracker'"
       end
       text = text.strip
+      text = "<span>#{text}</span>" if attributes =~ /start/
       href = href.strip
 
       %(\n<a role="button" draggable="false" class="#{button_classes}" href="#{href}" #{data_attribute}>#{text}</a>\n)
--- a/test/govspeak_button_test.rb
+++ b/test/govspeak_button_test.rb
@@ -118,7 +118,7 @@ class GovspeakTest < Minitest::Test
   " do
     assert_text_output "Start now JSA"
     assert_html_output %(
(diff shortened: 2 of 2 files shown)


CristinaRO:
The first link paste goes to somewhere unrelated, did you mean to link to the button partial?
https://github.com/alphagov/govuk_publishing_components/blob/9f51c7c150b81299ffda00a10c686a63a294e6c1/app/views/govuk_publishing_components/components/_button.html.erb#L18

Would that be the correct place to insert a `span` around the existing `button.text`?

kevindew:
> The first link paste goes to somewhere unrelated, did you mean to link to the button partial? https://github.com/alphagov/govuk_publishing_components/blob/9f51c7c150b81299ffda00a10c686a63a294e6c1/app/views/govuk_publishing_components/components/_button.html.erb#L18

🤦 sorry about that, multitasking and then had the wrong item on my clipboard. Yup that's what I meant to link to. Thanks

> Would that be the correct place to insert a `span` around the existing `button.text`?

Yup. I'd imagine the best place would be in this block: https://github.com/alphagov/govuk_publishing_components/blob/9f51c7c150b81299ffda00a10c686a63a294e6c1/app/views/govuk_publishing_components/components/_button.html.erb#L10-L15C10 that sets HTML for the start button

CristinaRO:
Moved to https://github.com/alphagov/govuk_publishing_components/pull/3478

PR #279, opened 2023-07-04
Disallow usage of style element and style attribute

kevindew:
This forbids a user from inputting a style element as part of their
govspeak markdown or using the style attribute in their HTML.
Thus it prevents publishers from customising the style of a GOV.UK page.

I am very confident we never intended to let users do this and consider
this a bug fix - luckily, the evidence of usage is minimal. This bug
provides publishers with an unprecedented ability to re-style a GOV.UK
page in a way far outside of the GOV.UK brand guidelines.

The reason I'm confident this was not intended is because we went to
special effort to only style on table cell elements 7 years ago [1]. It
appears that when we adopted Sanitize::Config::Relaxed in 2012 [2],
style was not allowed as it was later added to Sanitize::Config::Relaxed
in 2014 [3].

The motivation for making this change is that, with the [GOV.UK Content
Security Policy](https://github.com/alphagov/govuk_app_config/pull/291)
style is blocked, so even if users could use this feature it wouldn't change
the frontend (but would generate reports).

I've been working through the apps to make sure publishers will only
have minimal effect from this change and know there's a couple of docs
in Whitehall that need fixing which will be co-ordinated with this gem release.

### DIFF
--- a/lib/govspeak/html_sanitizer.rb
+++ b/lib/govspeak/html_sanitizer.rb
@@ -71,8 +71,8 @@ class Govspeak::HtmlSanitizer
         # We purposefully disable style attributes which Sanitize::Config::RELAXED allows
         :all => Sanitize::Config::RELAXED[:attributes][:all] + %w[role aria-label] - %w[style],
         "a" => Sanitize::Config::RELAXED[:attributes]["a"] + [:data] + %w[draggable],
-        "svg" => Sanitize::Config::RELAXED[:attributes][:all] + %w[xmlns width height viewbox focusable],
-        "path" => Sanitize::Config::RELAXED[:attributes][:all] + %w[fill d],
+        "svg" => %w[xmlns width height viewbox focusable],
+        "path" => %w[fill d],
         "div" => [:data],
         # The style attributes are permitted here just for the ones Kramdown for table alignment
         # we replace them in a post processor.
--- a/lib/govspeak/html_sanitizer.rb
+++ b/lib/govspeak/html_sanitizer.rb
@@ -74,8 +74,8 @@ class Govspeak::HtmlSanitizer
         "svg" => Sanitize::Config::RELAXED[:attributes][:all] + %w[xmlns width height viewbox focusable],
         "path" => Sanitize::Config::RELAXED[:attributes][:all] + %w[fill d],
         "div" => [:data],
-        # @TODO  These style attributes can be removed once we've checked there
-        # isn't hardcoded HTML in documents that uses them
+        # The style attributes are permitted here just for the ones Kramdown for table alignment
+        # we replace them in a post processor.
         "th" => Sanitize::Config::RELAXED[:attributes]["th"] + %w[style],
         "td" => Sanitize::Config::RELAXED[:attributes]["td"] + %w[style],
         "govspeak-embed-attachment" => %w[content-id],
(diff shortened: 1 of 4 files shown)


lib/govspeak/html_sanitizer.rb
```diff
@@ -60,17 +60,22 @@ def sanitize(allowed_elements: [])
...
-        "path" => Sanitize::Config::RELAXED[:attributes][:all] + %w[fill d],
+        "svg" => %w[xmlns width height viewbox focusable],
+        "path" => %w[fill d],
         "div" => [:data],
-        # @TODO  These style attributes can be removed once we've checked there
-        # isn't hardcoded HTML in documents that uses them
+        # The style attributes are permitted here just for the ones Kramdown for table alignment
+        # we replace them in a post processor.
         "th" => Sanitize::Config::RELAXED[:attributes]["th"] + %w[style],
         "td" => Sanitize::Config::RELAXED[:attributes]["td"] + %w[style],
```
ollietreend, line 80: Am I right in thinking this means a user could still provide arbitrary styles on `<th>` and `<td>` elements? And are we okay with that?

For example, would this work?

```markdown
## Some heading

<table>
  <tr>
    <td style="font-family: 'Comic Sans MS'; color: pink;">
      Will this work?
    </td>
  </tr>
</table>
```

Handily, it looks like the Sanitize gem allows us sanitize CSS properties too. The defaults provided by [Sanitize::Config::RELAXED](https://github.com/rgrove/sanitize/blob/a92f21cd223a32a1737262d68e56a4fb8b9470f9/lib/sanitize/config/relaxed.rb#L39-L721) are quite permissive – so perhaps we'd want to lock this down to just `text-align`?

e.g.

```ruby
css: {
  properties: %w[text-align]
}

File: test/html_sanitizer_test.rb [hunk 2]
```diff
@@ -17,6 +17,16 @@ class HtmlSanitizerTest < Minitest::Test
     assert_equal "<a href=\"/\">Link</a>", Govspeak::HtmlSanitizer.new(html).sanitize
   end
 
+  test "disallow style attributes" do
+    html = '<a href="/" style="font-weight:bold">Link</a>'
+    assert_equal "<a href=\"/\">Link</a>", Govspeak::HtmlSanitizer.new(html).sanitize
```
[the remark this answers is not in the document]
kevindew: Good shout, resolved.

lib/govspeak/html_sanitizer.rb
kevindew: No, it's not that arbitrary styles are allowed on those elements. There is some pre-existing code (see below) that removes the style attribute of td and th elements if they have a style value that isn't text-align: `(left|center|right)`. This is what convinced me that the style ability was a bug as someone had carefully set-up regex's to restrict access.
https://github.com/alphagov/govspeak/blob/907314932068b68f18e6d8020fb03d35700241f4/lib/govspeak/html_sanitizer.rb#L20-L36, 

It's a good shout about the sanitize config though, I wonder if that would be cleaner than the ruby we have - I'll have a quick experiment.

kevindew: I've had a look and can apply it. It does clean things up, but I ended up with a bit of churn along the way. I think it's still a net improvement. Let me know what you think.

ollietreend: @kevindew nice! 🎉  I definitely think this simplifies the implementation. Thanks!

ollietreend (approved):
Looks good to me 👍🏻

PR #283, opened 2023-08-01
Allow inline attachments in numbered lists

brucebolt:
At the moment, inline attachments do not work if included in a numbered list.

Adding a fix to allow this, as expected by publishers.

See commit messages for implementation details.

After merging and updating the gem in Whitehall, the affected document will need to be republished.

[Trello card](https://trello.com/c/37idNKRb)


JonathanHallam (approved):
nice! Thanks :)

PR #285, opened 2023-08-21
Reintroduce support for acronyms in Legislative Lists and Calls To Action

brucebolt:
This adds support for acronyms in Legislative Lists and Calls to Action, which was removed in https://github.com/alphagov/govspeak/pull/229.

The removal was because acronyms that matched parts of the HTML tags/attributes were also being replaced with acronym markup.

Therefore resolving this by searching for acronyms before converting to HTML.

[Trello card](https://trello.com/c/zzYVzGFM)


CristinaRO (approved):
Thank you for the detailed commit messages and step-by-step walkthrough! I think I understand enough to approve, but I won't be offended if you want to get a second opinion from someone with more experience in this area.

PR #293, opened 2023-10-18
Fix issues with block elements inside CTA and LegislativeList

jkempster34:
https://trello.com/c/VyfTCVe1

In https://github.com/alphagov/govspeak/pull/291, we made a change to how the call to action and legislative list
components are parsed. This involved the use of the `parse_block_html` option,
which enables parsing in HTML blocks until it is toggled off.

Since this change, we have received Zendesk tickets regarding the use of the
image component inside a call to action. Because we toggle off the parsing of
HTML blocks entirely within call to actions, the HTML of the image is not
parsed correctly.

To fix this, we can enable parsing of the surrounding divs of the call to
action and legislative list components only. This means that any inner blocks
(such as the `figure` tag surrounding an image) are parsed using their default
mechanism [1].

[1]: https://kramdown.gettalong.org/syntax.html#html-blocks.

## Before
image

## After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.3.1
+
+* Bug fixes related to block elements in call to action and legislative list components ([#293](https://github.com/alphagov/govspeak/pull/293))
+
 ## 8.3.0
 
 * Various bug fixes related to abbreviations in call to action and legislative list components ([#291](https://github.com/alphagov/govspeak/pull/291))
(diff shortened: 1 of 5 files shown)


mike29736 (approved):
I can't say I fully understand the different outcomes I'd expect from the different settings of that `markdown` attribute. But the tests are happy and the Govspeak converter examples look good, so it's going in the first direction :+1:

PR #298, opened 2023-12-12
Remove Kramdown overrrides

jkempster34:
https://trello.com/c/lu30wpdq

Currently, in order to disable ordered lists for legislative lists, we modify the @@parsers class variable and the LIST_START constant of the Kramdown library.

We are seeing a bunch of errors relating to these missing in production, such as:
- NoMethodError: undefined method `start_re' for nil:NilClass (NoMethodError) if @src.check(@parsers[name].start_re) 
- Kramdown::Error: Unknown parser: list (Kramdown::Error)
- NameError: uninitialized constant Kramdown::Parser::Kramdown::LIST_START

Rather than modifying class variables and constants in a non thread-safe manner, this takes the simpler approch of copying the body of the `parse_list` method and making the neccessary changes required for disabling ordered lists.

The downside of copying the majority body of the `parse_list` method may be offset by the fact that:
- This is thread-safe
- This is significantly simpler to understand
- The list module of Kramdown is rarely updated.
- The previous implementation also relied on the internals of Kramdown.

---


lib/kramdown/parser/govuk.rb
```diff
@@ -53,17 +53,110 @@ def add_link(element, href, title, alt_text = nil, ial = nil)
       def parse_block_html
         return false if CUSTOM_INLINE_ELEMENTS.include?(@src[1].downcase)
 
+        super
+      end
+
+      # We override the parse_list method with a modified version where ordered lists are
+      # disabled (for use with legislative lists). The majority of the method body is copied
+      # over, only the LIST_START is changed. Previously, we dynamically modified some of the
+      # class-scoped variables used in this method; this provides a thread-safe alternative.
```
mike29736, line 62: How would you feel about including a link to Github for the version of the method you copied in this comment? (e.g. maybe it was this one https://github.com/gettalong/kramdown/blob/REL_2_3_1/lib/kramdown/parser/kramdown/list.rb#L54 ?)

mike29736 (approved):
Personally, I think this is cleaner than the previous approach.

Considering that we haven't managed to reproduce the errors (and that we possibly wouldn't even want to reproduce them in the test suite if it is a thread safety issue), are you confident that the existing tests cover this area as well as they could?

Testing a branch of the gem in Whitehall sounds workable. But if you're happy with this, we could just release it generally and monitor it?

jkempster34:
> Personally, I think this is cleaner than the previous approach.
> 
> Considering that we haven't managed to reproduce the errors (and that we possibly wouldn't even want to reproduce them in the test suite if it is a thread safety issue), are you confident that the existing tests cover this area as well as they could?
> 
> Testing a branch of the gem in Whitehall sounds workable. But if you're happy with this, we could just release it generally and monitor it?

I've pulled the Govspeak out of the content items which are raising these errors, and all of them work as expected in Govspeak Preview. I think the test suite is fairly comprehensive when it comes to legislative lists. What would you recommend we call the release?

lib/kramdown/parser/govuk.rb
jkempster34: Done 👍

PR #302, opened 2024-01-24
Fix single line formatting of call to action component

jkempster34:
https://trello.com/c/iRNwQY9I

We use the default mechanism for parsing the call to action component so that
the parser handles abbreviations and footnotes within.

Since this change (#292) we have seen a regression where defining a call to
action on a single line does not work as expected (e.g., `$CTA Click here to
start the program $CTA`). The closing `div` is evaluated as markdown and the
call to action component therefore runs on until the next closing `div`.

This can be fixed by adding in some new lines between the div tags.

|Before recent changes (version 8.2.1)|Current (version 8.3.2)|This PR|
|-|-|-|
|image|image|image|
|image|image|image|
|image|image|image|
|image|image|image|

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fix single line formatting of call to action component ([#302](https://github.com/alphagov/govspeak/pull/302))
+
 ## 8.3.2
 
 * Various bug fixes related to legislative list components ([#298](https://github.com/alphagov/govspeak/pull/298))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fix single line formatting of call to action component
```
brucebolt, line 3: Can you add a link back to this PR.

brucebolt (approved):

PR #350, opened 2024-10-03
Support acronyms within example and address blocks

brucebolt:
Fixes an issue where acronyms are not correctly marked up when within example or address blocks.

[Trello card](https://trello.com/c/Eq8xjQV7)

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -302,7 +302,7 @@ module Govspeak
     end
 
     extension("address", surrounded_by("$A")) do |body|
-      %(\n<div class="address"><div class="adr org fn"><p>\n#{body.sub("\n", '').gsub("\n", '<br />')}\n</p></div></div>\n)
+      %(\n<div class="address"><div class="adr org fn"><p markdown="1">\n#{body.sub("\n", '').gsub("\n", '<br />')}\n</p></div></div>\n)
     end
 
     extension("legislative list", /#{NEW_PARAGRAPH_LOOKBEHIND}\$LegislativeList\s*$(.*?)\$EndLegislativeList/m) do |body|
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -296,7 +296,7 @@ module Govspeak
     extension("example", surrounded_by("$E")) do |body|
       <<~BODY
         <div class="example" markdown="1">
-          #{body.strip}
+          #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
         </div>
       BODY
     end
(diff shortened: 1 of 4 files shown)


lib/govspeak.rb
```diff
@@ -292,10 +292,17 @@
...
+      <<~BODY
+        <div class="example" markdown="1">
+          #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
+        </div>
+      BODY
+    end
 
     extension("address", surrounded_by("$A")) do |body|
-      %(\n<div class="address"><div class="adr org fn"><p>\n#{body.sub("\n", '').gsub("\n", '<br />')}\n</p></div></div>\n)
+      %(\n<div class="address"><div class="adr org fn"><p markdown="1">\n#{body.sub("\n", '').gsub("\n", '<br />')}\n</p></div></div>\n)
```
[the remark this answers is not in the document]
brucebolt: This is unchanged code.  It appears we want to replace the first new-line with nothing, then the subsequent ones with `<br />`.

lib/govspeak.rb
```diff
@@ -292,7 +292,14 @@ def render_image(image)
     wrap_with_div("place", "$P", Govspeak::Document)
     wrap_with_div("information", "$I", Govspeak::Document)
     wrap_with_div("additional-information", "$AI")
-    wrap_with_div("example", "$E", Govspeak::Document)
+
+    extension("example", surrounded_by("$E")) do |body|
+      <<~BODY
+        <div class="example" markdown="1">
+          #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
```
yndajas, line 299: **suggestion (if-minor):** documentation

Is this some kind of default? I can see that it's the same as what's done with CTAs, but it's different to how the address blocks code looks. Is there code or other documentation that could be linked to in this and maybe the next commit message to explain why:
1. moving from `wrap_with_div ... Govspeak::Document` to this RegEx is suitable here?
2. `markdown="1"` seems to be the key change for the abbreviation substitution to kick in (based on the next commit)

test/govspeak_test.rb
```diff
@@ -1247,6 +1247,20 @@ class GovspeakTest < Minitest::Test
     )
     end
 
+  test_given_govspeak "
```
yndajas, line 1263: **thought:** documentation

Would be nice if this method took some kind of human-readable description argument, but now's probably not the time to go adding that to this massive spec!

yndajas (approved):
Like George, I'm not super familiar with the syntax here, but the tests and their similarity to other acronym tests give me sufficient confidence this does the job

---

The labels/first lines of each of the comments here are following the [Conventional Comments](https://conventionalcomments.org/) format, which I'm trying out after watching [a useful DevOpsDays London talk](https://www.youtube.com/watch?v=5TnNlfPU7kY)

test/govspeak_test.rb
brucebolt: It would be nice to break this huge file down into some smaller ones, or even just wrap some context blocks around things.  I'll add a card to the backlog so it's not forgotten, as I don't think it'll be a trivial task to understand what exactly is being tested in each example.

PR #355, opened 2024-10-21
 (647) Support Embeds in Govspeak

pezholio:
This moves the logic to extract embedded content information and render embeds from the Publishing API into the Govspeak code. This gives us the flexibility to rub embed logic in other places within the publishing estate (for example: inline previews within Whitehall), as well as providing a single source of truth for supported content types without having to call to the Publishing API.

Assuming we all agree that this is the way forward, the next step will be to use this code in the Publishing API, as well as write up an ADR to record this decision. We will also then be able to use this code to allow embeds to work in the live preview in Whitehall.


README.md
```diff
@@ -671,6 +699,7 @@ which outputs
 </a>
 ```
 
+
```
brucebolt, line 702: Is this extra line intentional?

brucebolt (approved):
This approach makes sense to me.  Removing the parsing logic from Publishing API will also make the embeds more consistent with other types of markdown.

Could you add an entry to the `Unreleased` section of the `CHANGELOG` file before merging this PR (or bump the gem version is this PR, including the `CHANGELOG` entry).

README.md
pezholio: Whoops, sorted!

pezholio:
Nice one, I'll get on moving the functionality in Publishing API (with an ADR) tomorrow 👍

PR #356, opened 2024-10-23
Ignore embed codes when argument is blank

pezholio:
After pushing out https://github.com/alphagov/govspeak/pull/355, I realised that there are some circumstances (such as in Whitehall), where Govspeak is rendered before sending to the Publishing API. This results in embed code being stripped out before making its way to the Publishing API.

Rather than yanking the version, or reverting the code, I've changed the logic to return the embed code, rather than returning an empty string if the `content_blocks` argument is blank.

I've also renamed some of the content block related code to be more expressive of what is being embedded (there are other types of embeds elsewhere). This could be considered a breaking change, but considering I'm the only person using this code, I think it's fine.

It might be prudent to yank version `8.5.0` once this is fully rolled out though.

### DIFF
--- a/test/govspeak_content_blocks_test.rb
+++ b/test/govspeak_content_blocks_test.rb
@@ -42,12 +42,26 @@ class GovspeakContentBlocksTest < Minitest::Test
     assert_equal compress_html(expected), compress_html(rendered)
   end
 
-  it "ignores missing embeds" do
+  it "removes embed code if a content block cannot be found" do
+    content_block = {
+      content_id: SecureRandom.uuid,
+      document_type: "contact",
+      title: "foo",
+    }
+
     govspeak = "{{embed:contact:#{content_id}}}"
 
-    rendered = Govspeak::Document.new(govspeak, content_blocks: []).to_html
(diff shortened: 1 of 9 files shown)


test/govspeak_content_blocks_test.rb
```diff
@@ -42,14 +42,28 @@ def compress_html(html)
...
+
     govspeak = "{{embed:contact:#{content_id}}}"
 
-    rendered = Govspeak::Document.new(govspeak, content_blocks: []).to_html
+    rendered = Govspeak::Document.new(govspeak, content_blocks: [content_block]).to_html
 
     assert_equal compress_html(""), compress_html(rendered)
   end
 
+  it "ignores an embed code if content_blocks are not specified" do
```
yndajas, line 59: **suggestion:** test description

By "ignores" do you mean "retains"?

yndajas (approved):

test/govspeak_content_blocks_test.rb
pezholio: Good catch!

PR #357, opened 2024-10-29
Remove Advisory component

catalinailie:
The advisory component has been earmarked for removal. It does not pull in styles from the design system and so has to be manually updated if it needs changing.
Its uses will be replaced with [information callouts](https://github.com/alphagov/govspeak?tab=readme-ov-file#information-callouts).

Trello card: https://trello.com/c/Z26fvy04/2957-work-out-how-to-retire-advisory-from-govspeak-gem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Remove Advisory component ([#357](https://github.com/alphagov/govspeak/pull/357))
+
 ## 8.8.3
 
 * Update dependencies (actionview < 8.0.2, sanitize < 8, rubocop-govuk = 5.0.7) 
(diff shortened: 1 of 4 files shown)


JonathanHallam (commented):
Really happy with the changes. Before we merge this though can we get an idea of how many documents are using the advisory extension because they'll need updating

beccapearce:
To summarise our slack conversation for ease of following, there are around 30 published editions that use the advisory component. We have a rake task we are planning to run in [whitehall](https://github.com/alphagov/whitehall/pull/9674) to remove them all. We will then check again to see if there are any instances of advisory we have missed before we merge in this PR and the related [PR in govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/4349)

beccapearce (approved):

PR #362, opened 2024-11-12
Use a custom Kramdown parser for example blocks

brucebolt:
This makes govspeak's processing of example blocks consistent with the kramdown gem, rather than using our own concept of extensions.

[Trello card](https://trello.com/c/Xktkan9L)


brucebolt:
Closing as we're not going down this route for now.

PR #364, opened 2024-11-18
Allow data attributes in span tags

pezholio:
The work we’re doing in Content Modelling means we will be rendering content blocks within span tags with data attributes, which will make identifying content blocks easier. As we may be re-presenting content block embeds to Govspeak, we need to ensure that these tags don’t get stripped out, so we need to add `data-*` attributes to the allowlist for span tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-# 8.6.1
+## 8.7.0
+
+* Allow data attributes in spans ([#364](https://github.com/alphagov/govspeak/pull/364))
+
+## 8.6.1
 
 * Update dependencies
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# 8.6.1
+# 8.7.1
```
brucebolt, line 1: Should this be 8.7.0?

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# 8.6.1
+# 8.7.0
```
brucebolt, line 1: Could you make this a h2 (`##`) so it's consistent with the other entries?

brucebolt (approved):

CHANGELOG.md
pezholio: Done! 🎉

PR #371, opened 2024-12-17
Strip trailing space from lines in address blocks

mike3985:
There's a feature in the Kramdown parser (i.e. the library that the
Govspeak parser is built on top of) where two or more trailing
spaces at the end of a line of Govspeak will be converted into an
additional line break in the output HTML.

Until recently, the contents of address blocks weren't "reparsed" in the
Govspeak parser, meaning that they weren't subject to this Kramdown
feature. But the Govspeak parser code was recently updated and the
contents of address blocks are now reparsed. This has resulted in
unintentional additional line spacing being made visible in some
addresses that have been rendered by this new version Govspeak.

A typical example of the issue looks like:
```
| Name
|
| Street
|
| Town
|
| Postcode
|
```

where it would've previously looked like:
```
| Name
| Street
| Town
| Postcode
```

By stripping the trailing space from Govspeak lines within address
blocks, we effectively disable the Kramdown feature and prevent
unintentional line breaks in addresses going forward.

BTW, it's still straightforward to introduce additional line breaks
deliberately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Strip trailing backslashes from the lines of an address
 * When replacing line breaks with "&lt;br&gt;"s in the parsing of address blocks,
 replace the whole "\r\n" line break and not just the "\n" character
+* Only strip a leading line break from an address block, not just any old first
+occurrence of a line break
 
 ## 8.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 replace the whole "\r\n" line break and not just the "\n" character
 * Only strip a leading line break from an address block, not just any old first
 occurrence of a line break
+* Strip trailing whitespace from the lines of an address
 
 ## 8.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+* Strip trailing backslashes from the lines of an address
+* When replacing line breaks with "&lt;br&gt;"s in the parsing of address blocks,
+replace the whole "\r\n" line break and not just the "\n" character
+
 ## 8.7.0
 
 * Allow data attributes in spans ([#364](https://github.com/alphagov/govspeak/pull/364))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
 # Changelog
 
-## Unreleased
+## 8.8.0
 
-* Strip trailing backslashes from the lines of an address
+* Strip trailing backslashes from the lines of an address ([#371](https://github.com/alphagov/govspeak/pull/371))
 * When replacing line breaks with "&lt;br&gt;"s in the parsing of address blocks,
-replace the whole "\r\n" line break and not just the "\n" character
+replace the whole "\r\n" line break and not just the "\n" character ([#371](https://github.com/alphagov/govspeak/pull/371))
 * Only strip a leading line break from an address block, not just any old first
-occurrence of a line break
-* Strip trailing whitespace from the lines of an address
+occurrence of a line break ([#371](https://github.com/alphagov/govspeak/pull/371))
+* Strip trailing whitespace from the lines of an address ([#371](https://github.com/alphagov/govspeak/pull/371))
 
 ## 8.7.0
 
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -305,7 +305,7 @@ module Govspeak
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.sub(/\r?\n/, '').gsub(/\\*\r?\n/, '<br />')}
+        #{body.lstrip.gsub(/\\*\r?\n/, '<br />')}
         </p></div></div>
       BODY
     end
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -305,7 +305,7 @@ module Govspeak
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.sub("\n", '').gsub("\n", '<br />')}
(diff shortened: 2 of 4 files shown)


lib/govspeak.rb
```diff
@@ -302,7 +302,12 @@ def render_image(image)
     end
 
     extension("address", surrounded_by("$A")) do |body|
-      %(\n<div class="address"><div class="adr org fn"><p markdown="1">\n#{body.sub("\n", '').gsub("\n", '<br />')}\n</p></div></div>\n)
+      <<~BODY
```
yndajas, line 305: **praise:** heredocs

Thanks for the heredoc explainer. I didn't know they came with flavours

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Recognise "\r\n" line breaks in address blocks
```
yndajas, line 5: **suggestion:** verb

Is it worth saying what gets done to them here?

lib/govspeak.rb
```diff
@@ -305,7 +305,7 @@ def render_image(image)
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.sub(/\r?\n/, '').gsub(/\r?\n/, '<br />')}
+        #{body.lstrip.gsub(/\r?\n/, '<br />')}
```
yndajas: **thought:** TIL

`lstrip` 👀

lib/govspeak.rb
```diff
@@ -305,7 +305,7 @@ def render_image(image)
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.lstrip.gsub(/\r?\n/, '<br />')}
+        #{body.strip.gsub(/ *\r?\n/, '<br />')}
```
yndajas, line 308: **question:** syntax coverage

Do we also need to strip trailing slashes, or are we confident no one ever used them (in a way that's now not wanted)?

mike3985:
BTW, to test on Govspeak Preview locally, checkout the branch and have Govspeak Preview use the local version of the gem

```diff
-gem "govspeak"
+gem "govspeak", path: "../govspeak"
```

yndajas (approved):

lib/govspeak.rb
mike3985: Ah right, two trailing backslashes have the same effect as two trailing spaces. https://kramdown.gettalong.org/quickref.html#paragraphs:~:text=using%20two%20spaces-,or%20two%20backslashes,-at%20the%20end

I've checked my local (a few weeks out-of-date) Whitehall database and can't find any uses of trailing slashes inside address blocks. (FWIW, I did see them used inside $C contacts on my travels, though.)

I think I've unintentionally broken this Kramdown feature (for address blocks) with my changes. It seems to remove some of the slashes but not others. I hope it isn't that `\r?` 🙈 I'll take a look, because we either want it to work or, as you suggested, we should fully remove them.

mike3985: It turned out that a few commits on this branch broke the trailing backslashes. It was easier to just strip them in the end rather than getting them working again.

But I have had to introduce that change in the same commit as another change. I couldn't figure out how to do either without breaking the other 😵

lib/govspeak.rb
```diff
@@ -305,7 +305,7 @@ def render_image(image)
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.sub("\n", '').gsub("\n", '<br />')}
+        #{body.sub(/\r?\n/, '').gsub(/\\*\r?\n/, '<br />')}
```
yndajas: **nitpick:** slash regex

It has to be a multiple of two trailing backslashes (the documentation says 2+, but my testing suggested it needs to be a multiple of two), so you could do `(?:\\\\)*`, but I can't see why anyone would be leaving random backslashes on the ends of address block lines anyway

lib/govspeak.rb
```diff
@@ -305,7 +305,7 @@ def render_image(image)
       <<~BODY
 
         <div class="address"><div class="adr org fn"><p markdown="1">
-        #{body.lstrip.gsub(/\\*\r?\n/, '<br />')}
+        #{body.lstrip.sub(/[\s\\]*\z/, '').gsub(/[ \\]*\r?\n/, '<br />')}
```
yndajas, line 308: **question:** regex syntax

Any reason for using `\s` in the `sub` but ` ` in the `gsub`?

yndajas (approved):

lib/govspeak.rb
mike3985: I wasn't sure whether this would be useful to try to document. And if so, where to do that. I could just add something to the commit message?

The significant difference is the use of the `\z`. It targets the very end of the `body` string, matching a point after any final line break characters.

Obviously in the `gsub` we're matching the ends of all lines by explicitly looking for their line break characters. But according to the test suite, the last line of an address block's content won't necessarily have a trailing line break.

So, the `gsub` matches space characters + slashes preceding a line break, while the `sub` matches any space characters + slashes + line break characters + any other whitespace characters preceding the very end of the `body`.

lib/govspeak.rb
mike3985: Sorry, this is another small thing I should've mentioned. I couldn't get odd numbers of backslashes to play nice. They didn't just end up in the output HTML as is, they broke the encoding of the HTML line break 😓 That's why I went for the nuclear option

lib/govspeak.rb
yndajas: I was mostly curious about the use of a space vs `\s` than the other differences, which I think are easier to work out

mike3985: Ah right. I know I could've been more concise but I thought I was answering your question,

> Any reason for using `\s` in the `sub` but ` ` in the `gsub`?

Do you mean that this would've been a reasonable answer: "Because `\s` matches line break characters as well as space characters."?

yndajas: Ah, didn't know that!

PR #374, opened 2024-12-20
Restore behaviour of headings being able to be rendered inside example boxes

callumknights:
This is technically against govspeak usage guidance, but was behaviour that worked before, and the changing of implementation for example blocks in https://github.com/alphagov/govspeak/pull/350 broke it. 

Changing back to previous behaviour for now until more work is done on govspeak components and their interactions.

The strange behaviour with spacing appears to be related to this issue, credit to @brucebolt  https://github.com/gettalong/kramdown/issues/213 

Trello: https://trello.com/c/UVePLiKZ/1481-re-enable-headings-inside-example-boxes-in-govspeak

 ⚠️ This repo is owned by the publishing platform team. Please let us know in #govuk-publishing-platform when you raise any PRs. ⚠️

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -296,7 +296,7 @@ module Govspeak
     extension("example", surrounded_by("$E")) do |body|
       <<~BODY
         <div class="example" markdown="1">
-          #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
+        #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
         </div>
       BODY
     end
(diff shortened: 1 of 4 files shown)


lib/govspeak.rb
```diff
@@ -296,7 +296,7 @@ def render_image(image)
     extension("example", surrounded_by("$E")) do |body|
       <<~BODY
         <div class="example" markdown="1">
-          #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
+        #{body.strip.gsub(/\A^\|/, "\n|").gsub(/\|$\Z/, "|\n")}
```
JonathanHallam, line 299: This is a super good find. Would you mind just fixing the commit message over run issue you have and add a note in there about why this works please.

callumknights: Commit message fixed, and some explanation about the issue that Bruce found added above

brucebolt (approved):

PR #378, opened 2025-01-08
Release version 9.0.0

beccapearce:
Bump the version to 8.9.0
Changes included:
* Remove Advisory component ([#357](https://github.com/alphagov/govspeak/pull/357))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 9.0.0
 
 * Remove Advisory component ([#357](https://github.com/alphagov/govspeak/pull/357))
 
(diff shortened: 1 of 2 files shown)


yndajas (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 8.9.0
```
yndajas, line 3: **suggestion (non-blocking):** versioning

I see that the [reliant data should be removed](https://github.com/alphagov/govspeak/pull/357#issuecomment-2575282386), but I wonder if it's worth making this a major version bump anyway, since it's removing some functionality?

yndajas (approved):
Might be worth bumping this a major version, but given the data should already be removed, it's a non-blocking suggestion

PR #381, opened 2025-01-23
Fix CI (bumping minimum versions of Ruby and govuk_publishing_components)

mike3985:
https://trello.com/c/o6eA7900/1558-make-govspeak-tests-work-again


brucebolt (approved):
Nice approach!  This should save us a lot of time going forward.

PR #385, opened 2025-03-03
Add GA4 indexes to attachments that render a details component

AshGDS:
## What / Why
- Attachments that have an `alternative_format_contact_email` render a `details` component on them: https://components.publishing.service.gov.uk/component-guide/attachment/with_contact_email
- The performance analysts would like to be able to track which number each details element is on the page
- Therefore we need to set `index_section` (the index of the details component) and `index_section_count` (the total number of these attachments on the page)
- Therefore, I've written some code to achieve this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Add GA4 indexes to attachments that render a details component ([#385](https://github.com/alphagov/govspeak/pull/385))
+
 ## 10.0.1
 
 * Update dependencies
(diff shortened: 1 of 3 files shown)


brucebolt (approved):
Code change looks fine to me.  Could you add a `CHANGELOG` entry, under the `Unreleased` header then release the gem in another PR or another commit in this PR.

I did initially have some concerns about putting GA4 specific things into this gem (there's no prior art for this and the gem is [used outside of GDS](https://github.com/alphagov/govspeak/network/dependents)), but I think that's ok, since we've already got GA4 specific things in `govuk_publishing_components` so ultimately end up being output from govspeak anyway.

AshGDS:
Thanks @brucebolt - have updated the changelog and will do a release in a separate PR 👍 I wasn't aware this was used outside GDS, that's good to know!

PR #387, opened 2025-03-04
Allow acronyms/abbreviations in steps

brucebolt:
Abbreviations were not working when embedded in a step.

Fixing this in the same way as abbreviations in legislative lists: wrapping the list item in an element that has a `markdown="1"` value defined, then indenting the contents correctly so that the markdown is parsed again during the second round of processing.

Note: this affects the indentation, so the existing tests are being updated to reflect that.

[Trello card](https://trello.com/c/Pou7iXR4)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


callumknights (approved):
Looks good to me 🫠

PR #389, opened 2025-03-05
Revert "Add GA4 indexes to attachments that render a details component"

AshGDS:
This reverts commit ecdda77d0f39e115ae3b4d2a73f6fd0fd94e915f.

The code was working, but it isn't giving us exactly what we want - it was returning the count of all attachments for a given document, but we weren't aware that attachments can be added to a document but not rendered, or they can be rendered as an attachment link instead. We only want a count of attachments that are actually being rendered on the page, and are rendering as an `attachment` component. Therefore we'll explore using JS to handle this as I assume it'll be computationally expensive to try and get a total count of documents that are actually rendered as `attachment` components through `govspeak`, when we may be able to write some JS on the rendered page to group `attachment` components within the content and get the total count that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 10.2.0
+
+* Revert 'Add GA4 indexes to attachments that render a details component' ([#389](https://github.com/alphagov/govspeak/pull/389))
+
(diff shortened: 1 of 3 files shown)


yndajas (approved):
Thanks for cleaning it up!

PR #417, opened 2025-06-03
Improve localisation (particularly Welsh)

yndajas:
- Fix localisation for image credit label and add Welsh translation
- Test existing localisation of contact labels
- Add localisation of footnote labels


lib/govspeak/presenters/image_presenter.rb
```diff
@@ -1,9 +1,10 @@
 module Govspeak
   class ImagePresenter
-    attr_reader :image
+    attr_reader :image, :locale
 
-    def initialize(image)
+    def initialize(image, locale)
```
mike3985, line 5: This might be nicer as a named parameter?

mike3985 (approved):

lib/govspeak/presenters/image_presenter.rb
yndajas: Because we can assume a default for locale but image is _always_ variable?

PR #421, opened 2025-06-16
Support snapshot logging via an option

yndajas:
[Trello](https://trello.com/c/1iG6i2s2/1756-support-logging-in-govspeak)

This can be used in combination with a local copy of Govspeak Preview pointing at a local copy of Govspeak to test out changes

### DIFF
--- a/lib/govspeak.rb
+++ b/lib/govspeak.rb
@@ -37,7 +37,7 @@ module Govspeak
     @extensions = []
 
     attr_accessor :images
-    attr_reader :attachments, :contacts, :links, :locale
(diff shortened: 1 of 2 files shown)


lib/govspeak.rb
```diff
@@ -51,6 +51,9 @@ def initialize(source, options = {})
       options = options.dup.deep_symbolize_keys
       @source = source ? source.dup : ""
 
+      @log_snapshots = options.fetch(:log_snapshots, false)
+      log_snapshot("source and options", @source, options)
```
yndajas, line 55: **todo:** split this in two (from mobbing)

mike3985 (approved):
👍

PR #426, opened 2025-07-03
[PP-7224] Pass locale to nested calls

yndajas:
- Pass in a `Govspeak::Document`'s locale to any `Govspeak::Document`s it instantiates
- Use locale in `HtmlValidator` (essentially a nested call)


brucebolt (approved):
Is it worth adding something to the `CHANGELOG`, so this can be included in the next release?

PR #427, opened 2025-07-08
Remove legacy comments

brucebolt:
These `FIXME` and `TODO` comments have been present for 13 years without any action being taken.

There appears to be no action for us to take, nor any need for that to have been taken in those years, so removing them.

### DIFF
--- a/test/govspeak_test.rb
+++ b/test/govspeak_test.rb
@@ -341,8 +341,6 @@ Teston
 
   # Regression test - the surrounded_by helper doesn't require the closing x
   # so 'xaa' was getting picked up by the external link helper above
-  # TODO: review whether we should require closing symbols for these extensions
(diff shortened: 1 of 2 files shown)


test/govspeak_test.rb
```diff
@@ -341,8 +341,6 @@ class GovspeakTest < Minitest::Test
 
   # Regression test - the surrounded_by helper doesn't require the closing x
   # so 'xaa' was getting picked up by the external link helper above
-  # TODO: review whether we should require closing symbols for these extensions
```
JonathanHallam, line 344: should we do the todo?

brucebolt: Probably a bit late now as the extensions without closing symbols will be scattered throughout content.

PR #429, opened 2025-07-14
[PP-7224] Fall back to English on invalid locale or missing translation

yndajas:
We currently return an error string when provided with an invalid locale or a locale which doesn't have a required translation. Our support for locales has been expanded recently, so this might not be affecting any content (not much time for uptake), but we should address this so that it doesn't become an issue further down the line


JonathanHallam (approved):
This looks good to me! Thanks Ynda :)

PR #489, opened 2026-06-18
Add cards to govspeak

andysellick:
## What
Adds the [cards component](https://components.publishing.service.gov.uk/component-guide/cards) to govspeak.

Specifically, uses the cards component in `auto` mode, which makes the cards [display flexibly based on their container](https://components.publishing.service.gov.uk/component-guide/cards/auto_layout), with a `margin_bottom` of `6`.

Cards can be created in govspeak as follows:

```
{cards}
  [Benefits](https://www.gov.uk/browse/benefits) Includes eligibility, appeals, tax credits and Universal Credit
  [Births, deaths, marriages and care](https://www.gov.uk/browse/births-deaths-marriages) Parenting, civil partnerships, divorce and Lasting Power of Attorney
{/cards}
```

See included documentation for more examples.

## Why
Needed to reproduce the missions pages in a better format.

## Visual changes
Cards look like this:

Screenshot 2026-06-18 at 14 13 27

### DIFF
--- a/lib/govspeak/post_processor.rb
+++ b/lib/govspeak/post_processor.rb
@@ -147,6 +147,30 @@ module Govspeak
       end
     end
 
+    extension("use gem component for cards") do |document|
+      document.css(".cards").map do |cards|
+        links = cards.css(".card").map do |card|
+          {
+            link: {
+              path: card["href"],
+              text: card.css(".text").inner_html.html_safe,
+            },
+            description: card.css(".description").inner_html.html_safe,
+          }
+        end
+
+        cards_html = GovukPublishingComponents.render(
+          "govuk_publishing_components/components/cards", {
+            items: links,
+            columns: "auto",
+            margin_bottom: 6,
+          }
+        ).squish.gsub("> <", "><").gsub!(/\s+/, " ")
+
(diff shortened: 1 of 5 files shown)


lib/govspeak/post_processor.rb
```diff
@@ -147,6 +147,31 @@ def self.extension(title, &block)
       end
     end
 
+    extension("use gem component for cards") do |document|
+      document.css(".cards").map do |cards|
+        links = []
+        cards.css(".card").map do |card|
+          links << {
```
AshGDS: I think here you could just assign `links` to the `map` since a map returns an array anyway?

e.g. `links = cards.css(".card").map do |card|` and then just return each hash within the block instead of appending it with `links <<`

AshGDS (approved):
LGTM 👍

lib/govspeak/post_processor.rb
KludgeKML: Yeah, you don't need to use this empty array that you append to if you're using map.

andysellick: Thanks both, have updated.

KludgeKML (approved):
LGTM

PR #494, opened 2026-07-01
Revert "Add cards to govspeak"

andysellick:
Reverts alphagov/govspeak#489

This PR also includes an extra, new commit to create a new version of govspeak. In the original work this was done in two separate PRs.


AshGDS (approved):
LGTM 👍

